PR #75, opened 2025-12-23
Refactor frontend to KonstaUI elements

guillemcordoba:
This converts all pages of the app from using [WebAwesome elements](https://webawesome.com/) to [KonstaUI elements](https://konstaui.com/).


maackle (approved):
I skimmed, but it's good to know about

PR #81, opened 2026-01-14
Toy mailbox server

guillemcordoba:
- Watermark table for tracking contingent max sequence number stored.
- ToyMailboxClient implementation .
- Started to integrate into the tauri app.


crates/dashchat-node/src/mailbox/toy.rs
```diff
@@ -1,19 +1,157 @@
...
+        if ops.is_empty() {
+            return Ok(());
+        }
+
+        // Group operations by topic -> author -> seq_num
+        let mut blobs: BTreeMap<String, BTreeMap<String, BTreeMap<u64, Blob>>> = BTreeMap::new();
+
+        for op in ops {
+            let topic_id = log_id_to_topic_id(&op.topic());
+            let log_id = device_id_to_log_id(&op.author());
```
maackle: We should get clearer on the usage of the terms "topic", "log", "author", "device id".

My map of it is:
- topics and log ids are the same thing in our project, just two names used for two different contexts. Topics are for gossiping/syncing, Log IDs are for storing. p2panda lets you map many logs to a single topic, but we are currently mapping them 1:1
- a LogId (in the p2panda context) does not actually uniquely identify a log: you have to specify which author, too. Many authors can write their logs under the same log id. This is confusing to me but it's how p2panda does it.
- You know this already but a cryptographic "author" is always a device id.

So I think the mapping should be called:

    (topic | log_id) -> (author | device_id) -> seq_num

where we pick between those two options for the first two keys. I propose:

    topic -> author -> seq_num

but my main request is that we not use "log_id" to refer to the author/device_id

crates/mailbox-server/src/get_blobs.rs
```diff
@@ -0,0 +1,188 @@
...
+            }
+
+            // Check if this log was requested with a specific sequence number filter
+            let should_include = if let Some(min_seq_num) = requested_logs.get(&log_id) {
+                // Log is in the request: only include if seq_num > min_seq_num
+                seq_num > *min_seq_num
+            } else {
+                // Log is NOT in the request: include all blobs for this log
+                true
+            };
```
maackle: I wonder at what point we need to change this.

Maybe we can get away with a streaming response instead of full-on pagination? That would require more care around serialization but would help with performance. https://www.ruststepbystep.com/how-to-stream-data-asynchronously-in-rust-with-axum/

If that turns out not to be feasible we can put together some pagination by enforcing limits on how much data is sent back (ideally in bytes rather than in number of operations), and for each log (topic + author) the mailbox also specifies whether it reached the end of the log or not, and for each unfinished log, you make another request with the minimum sequence number increased to the new offset until all logs are complete.

We probably don't need to do anything right now but should make a note of our intention.

crates/mailbox-server/src/store_blobs.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+pub async fn store_blobs(
+    State(state): State<AppState>,
+    Json(payload): Json<StoreBlobsRequest>,
+) -> Result<StatusCode, (StatusCode, String)> {
+    let db = state.db.clone();
+    // Use spawn_blocking because redb's begin_write() is a blocking call that waits
+    // for exclusive write access. Running this directly in async context would block
+    // tokio worker threads and cause deadlocks under concurrent load.
```
maackle: Maybe for the "real" mailbox if there are lots of concurrent writes we will need to have a pool, behind a semaphore, but this seems good for the toy

crates/mailbox-server/src/watermark.rs
```diff
@@ -0,0 +1,128 @@
...
+    let mut sequences_per_log: BTreeMap<String, BTreeSet<SequenceNumber>> = BTreeMap::new();
+
+    {
+        let read_txn = db.begin_read()?;
+        let table = read_txn.open_table(BLOBS_TABLE)?;
+
+        // Note: redb's iter() returns (key, value) pairs. We only access the key.
+        // The value's AccessGuard is dropped without deserialization.
+        for entry in table.iter()? {
+            let (key, _value) = entry?;
```
maackle: Could be nice to be even more explicit about this?

```suggestion
            let (key, value) = entry?;
            drop(value);
```

maackle (commented):
Really nice!

PR #96, opened 2026-01-30
Enable sqlite persistance in dashchat-node, enable persistance in the development environment

guillemcordoba:
- Changed memory store to sqlite store in dashchat-node.
- Enable persistent development environment, storing databases in the `.dev-dbs` folder.
- Added `just` to make common development tasks easier to run.
- Added `just clean-dev` command to clean up development databases.

### DIFF
--- a/crates/dashchat-node/src/testing/test_node.rs
+++ b/crates/dashchat-node/src/testing/test_node.rs
@@ -11,8 +11,9 @@ use tokio::sync::{Mutex, mpsc::Receiver};
 use mailbox_client::{MailboxClient, mem::MemMailbox};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


crates/dashchat-node/src/node.rs
```diff
@@ -88,21 +90,24 @@ pub struct Node {
     /// Add new subscription streams
     stream_tx: mpsc::Sender<Pin<Box<dyn Stream<Item = Operation> + Send + 'static>>>,
 
+    filesystem: Filesystem,
```
maackle: this may only be used at init and maybe doesn't need to be stored?

PR #97, opened 2026-01-30
Local mailbox

maackle:
TODO: 

- [x] keepalive
- [x] consider autostart https://v2.tauri.app/plugin/autostart
- [x] ensure that a node can use its own mailbox server when enabled
- [x] only one mDNS-discovered address is registered, not all


src-tauri/src/mailbox.rs
```diff
@@ -0,0 +1,126 @@
...
+    });
+
+    Ok(())
+}
+
+fn mdns_service_info<R: Runtime>(_handle: &AppHandle<R>) -> ServiceInfo {
+    let ip = local_ip_address::local_ip().unwrap().to_string();
+    let instance_name = &nanoid::nanoid!();
+    let host_name = &format!("{ip}.local.");
+    let port = 3456;
```
maackle: @guillemcordoba I wonder if you have experience with mdns-sd and can help me configure these params to make mdns work. I'm trying this on two different computers on the same network and I've tried fiddling these params a lot to no avail.

The goal is to get line 94 to print

guillemcordoba (commented):
Looking really good! I was able to check out the branch, and confirm that the local mailbox mode is working as expected.

I left a few comments to iron out.

Also, before merging this we should consider its interaction with the [single-instance plugin](https://v2.tauri.app/plugin/single-instance/). We probably want to enable a single instance of the app running at any given time, otherwise we'll have two instances of all processes reading and writing from the same database files, and having two tray icons. I can take a look at that when everything else is ready, as I've done this before.

src-tauri/src/tray.rs
```diff
@@ -0,0 +1,141 @@
...
+        app.manage(mailbox::LocalMailboxMutex::new(None));
+        let mailbox_enabled = settings::load_mailbox_enabled(&handle);
+
+        if mailbox_enabled {
+            mailbox::start_local_mailbox(&handle)?;
+        }
+
+        let tray = crate::tray::build_tray(&app)?;
+        tray.set_visible(mailbox_enabled)?;
+        app.manage(tray);
```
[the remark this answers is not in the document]
guillemcordoba: We do get to pick the ID with https://docs.rs/tauri/latest/tauri/tray/struct.TrayIconBuilder.html#method.with_id, let's have be a hardcoded "DASH_CHAT_TRAY_ID" value.

src-tauri/src/tray.rs
```diff
@@ -0,0 +1,141 @@
+use tauri::tray::{TrayIcon, TrayIconBuilder};
+use tauri::{
+    menu::{Menu, MenuItem, PredefinedMenuItem},
+    App, AppHandle, Manager, Runtime,
+};
+
+pub fn build_tray<R: Runtime>(app: &App<R>) -> tauri::Result<TrayIcon<R>> {
+    let title = MenuItem::new(app, "DashChat Local Mailbox", false, None::<&str>)?;
+    let separator = PredefinedMenuItem::separator(app)?;
+    let show_i = MenuItem::with_id(app, "show", "Show", true, None::<&str>)?;
```
guillemcordoba: The menu and the tray both need to be localized. We can probably use https://crates.io/crates/sonix-i18n and include the strings from the ui folder

crates/mailbox-client/src/lib.rs
```diff
@@ -9,13 +9,18 @@ use std::{
     time::Duration,
 };
 
+use once_cell::sync::Lazy;
 use tokio::sync::{Mutex, mpsc};
 use tracing::Instrument;
 
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 
+pub static HTTP_CLIENT: Lazy<reqwest::Client> = Lazy::new(|| reqwest::Client::new());
```
guillemcordoba: Why did we change the client to be a static?

guillemcordoba (commented):
I wonder about the introduction of mailboxes IDs. I assume this is to detect that a local mailbox changed IP addresses but is still the same? I wonder if we actually need to introduce this tracking here, or we are doing it to start to go in the direction of tracking statistics for each mailbox. 

Let's imagine we don't track mailbox IDs. If a mailbox changes IP address and we just can't reach the old one, it will re-discover it in the new IP address but treat it as a new one, and everything should still work. Is there a problem I'm not seeing there?

PR #99, opened 2026-02-03
Reset contact code

guillemcordoba:
- Persist contact codes locally: Contact codes are now stored in the local redb database and reused across sessions, instead of generating a new code every time the user visits the add-contact page
- Add reset button with confirmation: Users can reset their contact code via a tonal button at the bottom of the page, with a confirmation dialog to prevent accidental resets
- Auto-regenerate expired codes: If a stored code has expired, a new one is silently generated
- Cleanup on reset: When resetting, the old inbox topic is removed from active inbox topics
- Rename QrCode → ContactCode: Renamed the struct to better reflect its purpose
- Add ContactCodeError enum: Extracted contact code errors into their own error type for better error handling


maackle:
Signal also doesn't allow users to query everyone who has used an active contact code, but our current implementation does, so that has to be taken into consideration for the UX too. As long as our infrastructure and security model isn't identical to Signal's, we can't assume that our UX will be.

Yes let's discuss

PR #109, opened 2026-02-23
Push notification server

maackle:
- Push notifications server (new crate) with SQLite storage, FCM integration, and automatic cleanup of invalid tokens.
- Push notifications client crate for shared types and HTTP client.
- Mailbox server triggers push notifications when new blobs arrive, with graceful shutdown support.
- Android: fully working — background push triggers mailbox sync, displays correct sender/message content, navigates to the right chat on tap, suppresses notifications for active chats.
- iOS: displays a generic "You have a new message" notification. 
  - TODO: add Notification Service Extension support to transform notifications with actual message content.
- Topic subscriptions synced at startup and forwarded in real time.
- Dev guide at docs/push-notifications.md.


crates/mailbox-server/src/store_blobs.rs
```diff
@@ -21,25 +21,56 @@ pub async fn store_blobs(
...
+                tracing::error!("Task join error: {}", e);
+                (StatusCode::INTERNAL_SERVER_ERROR, e.to_string())
+            })?
+            .map_err(|e| {
+                tracing::error!("{}", e);
+                (StatusCode::INTERNAL_SERVER_ERROR, e)
+            })?;
+
+    // Notify push notification subscribers for topics that received new data (non-blocking)
+    if !topics_with_new_blobs.is_empty() {
```
jadehopepunk: This section is a bit long, and goes into too much specific detail for a store_blobs function. Consider extracting out a function to enqueue push tasks

crates/mailbox-server/tests/push_integration.rs
```diff
@@ -0,0 +1,213 @@
...
+    let resp = mailbox.post("/blobs/store").json(&store_body).await;
+    resp.assert_status(StatusCode::CREATED);
+    tokio::time::sleep(std::time::Duration::from_millis(500)).await;
+
+    // Second store with same data — watermark won't advance, no push
+    let resp = mailbox.post("/blobs/store").json(&store_body).await;
+    resp.assert_status(StatusCode::CREATED);
+    tokio::time::sleep(std::time::Duration::from_millis(500)).await;
+
+    // MockFcm asserts exactly 1 call on drop
```
jadehopepunk, line 250: Nice tests :)

crates/push-notifications-client/src/client.rs
```diff
@@ -0,0 +1,115 @@
...
+            .send()
+            .await
+            .context("failed to send unsubscribe request")?
+            .error_for_status()
+            .context("unsubscribe request failed")?;
+
+        Ok(())
+    }
+
+    pub async fn notify_topics_request(
```
jadehopepunk: This method name is a bit unclear to me. Is it notifying of a request for topics, or is it requesting a topic notification.

crates/push-notifications-client/src/client.rs
```diff
@@ -0,0 +1,115 @@
...
+            .send()
+            .await
+            .context("failed to send notify-topic request")?
+            .error_for_status()
+            .context("notify-topic request failed")?;
+
+        Ok(())
+    }
+
+    pub async fn set_subscriptions(
```
jadehopepunk: Also a pretty strange name. Could it have a clearer verb? Maybe "subscribe_to_topics"?

crates/push-notifications-server/src/routes/notify_topic.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use crate::{AppState, error::AppError, fcm_client::SendResult};
+
+pub(crate) async fn notify_topics(
+    State(state): State<AppState>,
+    Json(req): Json<NotifyTopicsRequest>,
+) -> Result<StatusCode, AppError> {
+    let mut tasks = Vec::new();
+
+    for (topic_id, op_ids) in &req.topics_to_notify {
```
jadehopepunk: It might be worth extracting a notify_topic function, even if it's private. That will mean that people reading it don't need to think about the loop.

crates/push-notifications-server/src/routes/notify_topic.rs
```diff
@@ -0,0 +1,82 @@
...
+            Err(e) => {
+                tracing::warn!(topic_id = %topic_id, "failed to get subscribers: {e:#}");
+                continue;
+            }
+        };
+
+        // One push per operation: iOS Notification Service Extensions can only
+        // transform a single incoming notification, so each operation needs its own push
+        // for the client to resolve it into a user-facing message.
+        for op_id in op_ids {
```
jadehopepunk: I'd probably extract a method here too, notify_of_operation or something like that.

crates/push-notifications-server/src/fcm_client.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    async fn send_push_notification(
+        &self,
+        token: &str,
+        notification: &PushNotification,
+    ) -> SendResult {
+        let mut message = Message::default();
+
+        let mut map = HashMap::new();
```
jadehopepunk: This variable name is very bland. I know it's a map, what's it for?

jadehopepunk (commented):
Looks good, a few small readability nits but overall very happy wit the client server design, and with the tests.

I'd approve except that I think maybe you should look at the comments around naming those API functions as they might be hard to change later.

crates/push-notifications-client/src/client.rs
guillemcordoba: This was a leftover from a previous naming. Thanks!

crates/push-notifications-client/src/client.rs
guillemcordoba: Humm we already have `/subscribe`, which could be changed to `/subscribe-to-topics`. What I want this function to do as well, is remove the topics that have been subscribed to before that are not present in the payload for this request.

Having trouble coming up with a name... What do you think @jadehopepunk @maackle ?

crates/dashchat-node/src/node.rs
```diff
@@ -113,6 +113,7 @@ pub struct Node {
     // groups: p2panda_auth::group::Groups,
     config: NodeConfig,
     notification_tx: Option<mpsc::Sender<Notification>>,
+    topic_subscribed_tx: Option<mpsc::Sender<TopicId>>,
```
maackle, line 116: The codebase currently isn't well set up for unsubscribing to topics. Making a note to self here to consider that as I read

crates/mailbox-server/tests/push_integration.rs
```diff
@@ -0,0 +1,213 @@
...
+                        "0": blob_data
+                    }
+                }
+            }
+        }))
+        .await;
+    resp.assert_status(StatusCode::CREATED);
+
+    // Give the spawned push notification task time to complete
+    tokio::time::sleep(std::time::Duration::from_millis(500)).await;
```
maackle: Do you suppose these tests could wait on the JoinSet directly here? Dependence on wall time = flakyness.

crates/push-notifications-client/src/client.rs
maackle: When you describe what it does, "set" or "update" make sense, "update" maybe being a little better. Another option is "replace_subscriptions".

crates/push-notifications-client/src/requests.rs
```diff
@@ -0,0 +1,33 @@
...
+pub struct UnsubscribeRequest {
+    pub public_key: PublicKey,
+    pub topic_ids: HashSet<TopicId>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct SetSubscriptionsRequest {
+    pub public_key: PublicKey,
+    pub topic_ids: HashSet<TopicId>,
+}
```
maackle, line 146: It might make things clearer to unify these 3 in terms of nouns instead of some verbs some nounds:

- `AddSubscriptionsRequest`
- `RemoveSubscriptionsRequest`
- `{Set|Update|Replace|etc}SubscriptionsRequest`

maackle: I don't think this helps but in terms of set operations, it's:

- update with union of existing and provided set
- update with difference of existing and provided set
- replace with provided set

src-tauri/src/push_notifications/android.rs
```diff
@@ -0,0 +1,233 @@
...
+            Err(err) => {
+                log::error!("Failed to create node for push notification: {err:?}");
+                return None;
+            }
+        };
+
+        // Trigger a mailbox sync to fetch the new operation
+        node.mailboxes.trigger_sync();
+
+        // Poll for the operation to arrive (up to 15 seconds)
```
maackle, line 101: This seems fine for now, but how do we feel about documenting possible improvements that we may some day do? I like to do this kind of thing, even if it never gets done, it gives a starting point for improvements when the time is right.

```suggestion
        // Poll for the operation to arrive (up to 15 seconds)
        // PERF: consider adding the ability for the op store to notify when an op is stored,
        //     instead of polling
```

maackle (commented):
almost done reviewing, leaving comments so far, for now

src-tauri/src/push_notifications/android.rs
```diff
@@ -0,0 +1,233 @@
...
+        if sender_device_id == node.device_id() {
+            return None;
+        }
+
+        let body = body?;
+        let payload = match Payload::try_from_body(&body) {
+            Ok(p) => p,
+            Err(err) => {
+                log::error!("Failed to decode payload: {err:?}");
+                return Some(generic_notification());
```
maackle: Do we actually want to show a notification on deserialization error? That's more for our benefit but users won't want to see that.

Cargo.toml
```diff
@@ -11,19 +11,19 @@ resolver = "2"
...
+p2panda-net = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat-auth-processor" }
+p2panda-store = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat-auth-processor", features = [
   "sqlite",
 ] }
-p2panda-stream = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat" }
-p2panda-sync = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat" }
-p2panda-spaces = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat", features = [
+p2panda-stream = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat-auth-processor" }
+p2panda-sync = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat-auth-processor" }
+p2panda-spaces = { git = "https://github.com/maackle/p2panda.git", branch = "dashchat-auth-processor", features = [
```
maackle, line 26: Maybe I should merge this p2panda branch into "dashchat" first?

maackle (commented):
wowowow! Let's see how it works.

src-tauri/src/push_notifications/android.rs
guillemcordoba: I think you're right. The case in which I can see this happening is with version changes, a notification is pushed that is compatible with the version of the app the user has installed. In that case, even if we show a generic notification and the user taps it to open the app, the operation won't appear in the UI because it's not compatible, so the user will be left wondering what new thing actually happening. Changing so this returns None.

crates/mailbox-server/tests/push_integration.rs
guillemcordoba: Fixed!

crates/push-notifications-client/src/client.rs
guillemcordoba: Going with `/topic-subscriptions/add` and `add_topic_subscriptions()`,`/topic-subscriptions/remove` and `remove_topic_subscriptions()`,`/topic-subscriptions/update` and `update_topic_subscriptions()`

guillemcordoba:
These are the main changes from the state of the PR yesterday:

1. Notifications settings page — Added /settings/notifications route with a toggle to enable/disable push notifications, wired into the settings panel (mobile only).
2. Persistent notifications setting — Added notifications_enabled to the local settings system (Rust backend + frontend stores), defaulting to false until the user grants permission during profile creation. Backend gates FCM token registration and token refreshes on both this setting and OS permission state.
3. FCM token unregister flow — Added POST /fcm-tokens/unregister endpoint, called with retries when the user disables notifications
4. HTTP API rename — Renamed all push notifications server routes and types to be consistent: /subscribe → /topic-subscriptions/add, /unsubscribe → /topic-subscriptions/remove,  /set-subscriptions → /topic-subscriptions/update, /register-fcm-token → /fcm-tokens/register; request types and driver trait methods renamed to match.
5. Extracted exponential backoff utility — Pulled out a reusable retry_with_backoff function (in dashchat-utils) used across FCM token registration, subscription syncing, and unregistration to be resilient to network failures.

crates/dashchat-utils/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+use std::time::Duration;
+
+/// Retry an async operation with exponential backoff.
+///
+/// - `max_attempts: Some(n)` — give up after `n` attempts and return the last error.
+/// - `max_attempts: None` — retry forever until success.
+///
+/// Returns `Ok(T)` on the first successful attempt, or `Err(E)` after exhausting
+/// all attempts (only possible when `max_attempts` is `Some`).
+pub async fn retry_with_backoff<T, E, F, Fut>(
```
jadehopepunk, line 12: Nice, I like that this is a general utility method

jadehopepunk (approved):
My concerns are all addressed

PR #156, opened 2026-03-06
Add user preferences

jackaperkins:
Added here is a basic JSON-file backed preferences storage, with theme and language options sketched out.

As theme isn't well supported by the actual components yet the option isn't "wired up" to the DOM. Language selection works well using the preexisting paraglide setup.

Also added is the tauri backend support for deleting the local account, it simply deletes the databases and restarts tauri.


guillemcordoba:
Superseeded by https://github.com/dash-chat/dash-chat/pull/164

PR #157, opened 2026-03-06
Mailbox which handles blobs separately

maackle:
We may or may not want to move in this direction. (I believe we do, but we haven't talked about it yet.) This will allow clients to fetch and sync blobs separately from log headers, and to eventually separate blob storage into its own service (like a CDN) distinct from the service of syncing append-only log headers.

Note that the existing implementation unfortunately used "blobs" to refer to "operations" even though we use "blobs" elsewhere to mean the large opaque payloads of operations. This PR introduces mailboxes handling both headers and payloads. To the mailbox, these items are both opaque "blobs". To differentiate them, for now I named the opaque headers "dollops" (little blobs) and the opaque payloads "blobs" (big blobs). I also renamed the general notion of opaque bytes as used by the mailbox to be "Opaq" and "OpaqHash" (an Opaq can be a dollop or a blob).

This of course needs some serious rethinking of the naming, but now at least the names are non-colliding, and choosing new names for things are a relatively simple find-replace.

This is not high priority for us to look at but it's here whenever we want to.


guillemcordoba:
This should be closed as superseeded by https://github.com/dash-chat/dash-chat/pull/373 right @maackle ?

maackle:
superseded by https://github.com/dash-chat/dash-chat/pull/373

PR #158, opened 2026-03-07
Initial plumbing for Device groups

maackle:
This lays the groundwork for device groups, for linked devices.

The notable hacky thing here is that nested groups cannot be given Manage access, so in this PR if we are adding an AgentId as a Manager, we also add their DeviceId. This is OK because for now we don't actually allow linked devices, and every Agent only has one Device. This workout around is denoted by feature `auth-workaround`, and when p2panda implements this ability we can remove that feature and all code flagged by it.

Another thing to look out for is the handling of encountering new group members who I am not a contact of. In this PR, when I notice group operations for strangers, I automatically subscribe to their device group, because I cannot process those group operations until I process the device group which they depend on. We have to handle this better, including unsubscribing when I or anyone else leaves a group. I added a TODO about this.

There is a comprehensive test of 4 group members adding each other.

Question @guillemcordoba, will we support removals yet? Or even write access? Perhaps everyone is a manager to start, with no removals? If we support removals I will add tests for that too.

- [ ] test of members removing each other


crates/dashchat-node/src/stores/op_store.rs
```diff
@@ -265,7 +266,7 @@ impl OpStore<MemoryStore<TopicId, Extensions>> {
                     Some(p) => format!("{p:?}"),
                     None => "_".to_string(),
                 };
-                if topics.len() == 1 {
+                if false && topics.len() == 1 {
```
guillemcordoba: Put at least a comment about why false &&

crates/dashchat-node/src/node/stream_processing.rs
```diff
@@ -128,9 +128,36 @@ impl Node {
...
+                let members = match &auth.action {
+                    GroupAction::Create { initial_members } => initial_members.iter().map(|(member, _)| member).collect(),
+                    GroupAction::Add { member, .. } => vec![member],
+                    GroupAction::Remove { ..} => vec![],
+                    GroupAction::Promote { .. } => vec![],
+                    GroupAction::Demote { .. } => vec![],
+                };
+                for member in members {
+                    match member {
+                        GroupMember::Group(id) => self.register_topic(ChatId::from_group_pubkey(*id)).await?,
```
guillemcordoba: Shouldn't we also subscribe to their Announcements topic? Feels like some topic design cleanup might be in order. To me, it make sense that your "public key" topic is about announcing 1) your profile, 2) your linked devices. I don't understand why we are subscribing to a `ChatId` topic here?

guillemcordoba (commented):
@maackle okey... So this feels very hacky, specially three aspects of it:
- The fact that we need to add devices as their own full member.
- The fact that this disallows group member removal.
- The registering of DeviceGroup topics by other agents, when that topic is private in my mind.

I'm concerned that adding this to the codebase at this point makes it more brittle than necessary, with little added benefit in return. Can you make a case for why it is important to move to device groups at the moment and not in the future when all the right primitives are implemented, and no hacky workarounds are needed?

crates/dashchat-node/src/node/stream_processing.rs
maackle: You're right, this was a mistake. Semantically this needs to be the Announcements topic, and in fact that's what this code is doing, because all that matters is the underlying bytes, not the type -- Announcment topics happen to be represented by that agent key, so this actually worked just as well. But I am updating this to actually construct a Announcements topic so this is clearer, which I should have done in the first place.

crates/dashchat-node/src/stores/op_store.rs
maackle: This is basically test code so I got sloppy but yes, I'll clean that up

maackle:
> I'm concerned that adding this to the codebase at this point makes it more brittle than necessary, with little added benefit in return. Can you make a case for why it is important to move to device groups at the moment and not in the future when all the right primitives are implemented, and no hacky workarounds are needed?

I think it's important because this ensures that we're using AgentIds in the right way, and we can start validating now that this approach even works well. It forces us to align our patterns with this one. The longer we go without this pattern in the place, the more we may make bad assumptions, particularly about when to use DeviceIds vs AgentIds, and about the need for the mappings between them.

>    The fact that we need to add devices as their own full member.

Everything to do with the workaround is behind the feature flag `auth-workaround`, and when we remove that feature we will remove all code associated with it, which I think is a pretty clean way to keep this hack contained and easily removable. If you mentally subtract that feature flag now, I think you'll see little or no trace of the workaround.

>    The fact that this disallows group member removal.

This doesn't disallow removals -- I just didn't implement removal now. The only thing extra in implementing removal with this workaround is to make sure to remove the device as well as the agent. Most of the work doesn't have to do with the workaround -- like ensuring topics get unsubscribed to, maybe blocking removed members who persist in sending messages, ensuring that any group member cannot add arbitrary Individuals to a chat group, only Group members under AgentIds, preferably ensuring that the AgentId actually has a non-empty Announcements topic with device group operations, etc... some of this stuff would be better to wait for implementing until we have group encryption anyway.

>    The registering of DeviceGroup topics by other agents, when that topic is private in my mind.

This was a mistake which I flagged and fixed



We can discuss this workaround, and the timeline for removals

guillemcordoba:
Agreed! We can go ahead and continue to process this PR then. I merged develop but I'm finding problems with the current code, can you push your p2panda fork?

PR #160, opened 2026-03-10
Basic file and image support

maackle:
Needs plenty of UI polish! I tested this by hand not too extensively.

Mostly done by Claude. I will review too. Shall we talk about strategy for automated UI testing of this kind of feature?


crates/dashchat-node/src/chat/message.rs
```diff
@@ -2,22 +2,49 @@ use named_id::RenameNone;
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "kind")]
+pub enum MediaAttachment {
+    #[serde(rename = "photos")]
+    Photos { photos: Vec<PhotoAttachment> },
+    #[serde(rename = "file")]
+    File { file: FileAttachment },
+}
+
```
maackle, line 62: Does this seem right? a message can contain one file or many photos. I chose this intentionally but checking with you.

src-tauri/src/commands/redact_log.rs
```diff
@@ -19,8 +19,8 @@ static REDACTION_REGEXES: LazyLock<Vec<Regex>> = LazyLock::new(|| {
         r#""?timestamp"?\s*:?\s*\d{10,}"#,
         // Debug format: name/surname/about fields with quoted values
         r#"(name|surname|about):\s*(Some\()?"[^"]*"(\))?"#,
-        // Debug format: ChatMessageContent("...")
-        r#"ChatMessageContent\("[^"]*"\)"#,
+        // Debug format: ChatMessageContent { message: "...", ... }
+        r#"ChatMessageContent\s*\{[^}]*\}"#,
```
maackle, line 29: wasn't sure about the redaction stuff, double check this...

ui/src/lib/components/MessageInput.svelte
```diff
@@ -2,33 +2,94 @@
...
+	function removePhoto(index: number) {
+		if (!media || media.kind !== 'photos') return;
+		const remaining = media.photos.filter((_, i) => i !== index);
+		onMediaChange?.(remaining.length > 0 ? { kind: 'photos', photos: remaining } : undefined);
+		tick().then(updateHeight);
+	}
+
+	function updateHeight() {
+		if (div) height = `${div.scrollHeight}px`;
+	}
```
maackle, line 89: Do you suppose there's a less procedural way to do this?

ui/src/routes/direct-chats/[agentId]/+page.svelte
Comment (maackle) [inline ui/src/routes/direct-chats/[agentId]/+page.svelte:671]: we need to converge on the same file formatter!

crates/dashchat-node/src/chat/message.rs
```diff
@@ -2,22 +2,49 @@ use named_id::RenameNone;
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, RenameNone)]
+pub struct FileAttachment {
+    pub data: String,
+    pub name: String,
+    pub mime_type: String,
+    pub size: u64,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "kind")]
```
guillemcordoba: Let's use tag = "type", content = "payload", for consistency with the other enums in the app.

crates/dashchat-node/src/chat/message.rs
guillemcordoba: Same pattern as signal, sounds good!

crates/dashchat-node/src/chat/message.rs
```diff
@@ -2,22 +2,49 @@ use named_id::RenameNone;
...
+    File { file: FileAttachment },
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, RenameNone)]
+pub struct ChatMessageContent {
+    pub message: String,
+    pub media: Option<MediaAttachment>,
+}
+
+impl ChatMessageContent {
```
guillemcordoba, line 66: Interestingly enough, this represents our first breaking change in data structure. Let's actually take this opportunity to figure out a way to handle data structure breaking changes. The sooner we implement this the better, as the consequences compatiliby breakage for testers are very disruptive ("it deleted my whole profile, will it do that every time I upgrade?"). We want them to have as good an experience as possible.

I think there are different approaches, I like this one https://ivanbyte.medium.com/backward-compatible-data-de-serialization-with-serde-flow-in-rust-c87a2e8bc9ea

packages/stores/src/types.ts
```diff
@@ -23,7 +23,28 @@ export interface ChatReaction {
...
+
+export interface FileAttachment {
+	data: string;
+	name: string;
+	mime_type: string;
+	size: number;
+}
+
+export type MediaAttachment =
+	| { kind: 'photos'; photos: PhotoAttachment[] }
```
guillemcordoba: Rename to "type" and "payload"

ui/src/lib/components/MessageInput.svelte
guillemcordoba: No... This one was tricky to get right, basically the tracking of the height of the input needs to be "manual" to synchronize the size of the text rows with the outer div which is the one that carries the styles.

guillemcordoba (commented):
In general it looks good!

guillemcordoba:
@maackle I think this can be closed as superseeded by https://github.com/dash-chat/dash-chat/pull/374 right?

maackle:
superseded by #374

PR #161, opened 2026-03-10
Audio recording support

maackle:
Adds audio file support. Again, needs plenty of polish.

With this and the previous PR, we should consider switching away from data urls and actually use byte arrays. I used data urls here because they're used everywhere else too, for avatar images and such.

We also may choose a different codec. I chose mp3 because it is likely to be well supported in `<audio>` for all platforms, and has the native js implementation of lamejs.

### DIFF
--- a/src-tauri/src/tray.rs
+++ b/src-tauri/src/tray.rs
@@ -94,10 +94,24 @@ pub fn show_or_create_main_window<R: Runtime>(app: &AppHandle<R>) -> anyhow::Res
         window.show()?;
         window.set_focus()?;
     } else {
-        WebviewWindowBuilder::new(app, "main", WebviewUrl::default())
+        let window = WebviewWindowBuilder::new(app, "main", WebviewUrl::default())
             .title("Dash Chat")
             .inner_size(800.0, 600.0)
             .build()?;
+
+        #[cfg(target_os = "linux")]
+        {
+            window.with_webview(|wv| {
+                use webkit2gtk::{WebViewExt, PermissionRequestExt};
+                let webview: webkit2gtk::WebView = wv.inner().clone();
+                webview.connect_permission_request(|_wv, req| {
+                    req.allow();
+                    true
+                });
+            }).ok();
+        }
+        #[cfg(not(target_os = "linux"))]
+        let _ = &window;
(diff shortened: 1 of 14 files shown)


src-tauri/src/tray.rs
```diff
@@ -94,10 +94,24 @@ pub fn show_or_create_main_window<R: Runtime>(app: &AppHandle<R>) -> anyhow::Res
...
     } else {
-        WebviewWindowBuilder::new(app, "main", WebviewUrl::default())
+        let window = WebviewWindowBuilder::new(app, "main", WebviewUrl::default())
             .title("Dash Chat")
             .inner_size(800.0, 600.0)
             .build()?;
+
+        #[cfg(target_os = "linux")]
+        {
+            window.with_webview(|wv| {
```
guillemcordoba, line 104: Deduplicate with previous change

ui/src/lib/components/MessageInput.svelte
```diff
@@ -125,6 +139,74 @@
...
+			recordingTimer = setInterval(() => {
+				if (recorderHandle) {
+					recordingElapsed = Date.now() - recorderHandle.startTime;
+				}
+			}, 200);
+		} catch (err) {
+			if (err instanceof DOMException && (err.name === 'NotAllowedError' || err.name === 'NotFoundError')) {
+				showToast(m.microphonePermissionDenied());
+			} else {
+				console.error('Failed to start recording:', err);
```
guillemcordoba, line 164: We should show a toast of type unexpected here.

ui/src/lib/components/MessageInput.svelte
```diff
@@ -125,6 +139,74 @@
...
+				kind: 'audio',
+				dataUrl: result.dataUrl,
+				mimeType: result.mimeType,
+				durationMs: result.durationMs,
+				size: result.size,
+			});
+			await tick();
+			updateHeight();
+		} catch (err) {
+			console.error('Failed to stop recording:', err);
```
guillemcordoba, line 193: same, toast of type unexpected

guillemcordoba (commented):
Looking really good!

PR #166, opened 2026-03-12
Larger avatar in "Set Profile" page, like it is on signal desktop 

### LINKED ISSUE #125: Profile avatar should be bigger in "Set Profile" page, like it is on signal desktop

jadehopepunk:
Addresses: https://github.com/dash-chat/dash-chat/issues/125

Also, as in signal desktop, the avatar shouldn't change position when you move into edit mode, so changed the margin top to be the same as in the PhotoPicker component.

I don't love just having constants that match in two places. Some of this is pretty normal in these tailwind approaches though. I considered having a size param on the PhotoPicker that would get passed in, so that it could be the same constant, but this isn't actually the page that loads the PhotoPicker, so that seems a bit hard.


guillemcordoba:
Unfortunately I think the issue was not clear enough, it was supposed to be about the "Create Profile" page, rather than the "Profile Settings" page.

If I look at the [profile settings screenshot](https://github.com/dash-chat/signal-screenshots/blob/main/android/settings/profile/01-profile-overview.png), now our avatar definitely looks much bigger than theirs. 

And this issue was supposed to address the "Create Profile" page, which relates to [this screenshot](https://github.com/dash-chat/signal-screenshots/blob/main/android/create-account/15-profile-setup-name.png).

guillemcordoba (approved):

PR #168, opened 2026-03-17
Maintain avatar colour and text

### LINKED ISSUE #121: When reselecting the color choice tab, the text choice tab, the previous text and color choice dissappear.
1) in the avatar set page
2) select text option
3) go to text and color selection
4) select text and color
5) hit save and return to avatar selection
6) see text and color in avatar page
7) hit text button to return to avatar text and color choice page
8) previous text and color should be remembered and the starting point

jadehopepunk:
Fix for issue: https://github.com/dash-chat/dash-chat/issues/121


.vscode/settings.json
```diff
@@ -1,6 +1,94 @@
 {
 	"editor.defaultFormatter": "esbenp.prettier-vscode",
 	"[svelte]": {
-		"editor.defaultFormatter": "svelte.svelte-vscode"
```
jadehopepunk: This change doesn't relate to this PR really. It might not be needed. Are you using a plugin to automatically exclude this ignored files from VSCode's search and explorer? This change was generated by a plugin I'm using that does a one off addition of ignored file patterns to the exclude list.

ui/src/lib/components/profiles/Avatar.svelte
```diff
@@ -1,25 +1,43 @@
...
+	const avatarInitials = $derived(textAvatarData?.text ?? initials);
+	const avatarStyle = $derived.by(() => {
+		if (!textAvatarData) {
+			return style;
+		}
 
-	const profile = useReactivePromise(contactsStore.profiles, chatActorId);
+		const textAvatarStyle = `background-color: ${textAvatarData.color}; color: #831843;`;
+		return style ? `${style}; ${textAvatarStyle}` : textAvatarStyle;
+	});
```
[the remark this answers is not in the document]
jadehopepunk: Check again copilot, this now called sanitizedHexColor which performs the checking you suggest.

.vscode/settings.json
guillemcordoba: I think that's fine! I don't use .vscode, I'm happy to accept this as the correct vscode config.

jadehopepunk:
Ok @guillemcordoba this is ready for your review. 

At this point I don't agree with any of the further suggestions from Claude. 

Additionally, the issue we found while screensharing yesterday, where  you have a saved text image, then you go and pick an animal image, and you change your mind (without saving) and go and start a text image, and you don't have the original text image to start from. That also occurs in Signal (iOS). I think maybe leave it? It's hard to be sure what better behaviour would be. You can always cancel after choosing the animal image and then edit your text image.

guillemcordoba (approved):

PR #173, opened 2026-03-19
Ensure that name fields are not lost when opening/closing avatar picker

### LINKED ISSUE #120: Setting avatar AFTER putting name in, deletes the name input upon return to the main Set profile page
1) we're in the set profile page
2) enter name
3) navigate to avatar set page
4) select avatar
5) navigate back (name shouldn't dissappear)

jadehopepunk:
The fields were not controlled, and they were getting unmounted when the picker opened, now fixed.

As a larger consideration, coming from react, I would typically make a component like this more of a pure component, without talking to the state directly. Then, have the calling page do the handle the result of this save and writing it to the state. 

Fix for https://github.com/dash-chat/dash-chat/issues/120

### DIFF
--- a/ui/src/lib/components/profiles/CreateProfile.svelte
+++ b/ui/src/lib/components/profiles/CreateProfile.svelte
@@ -152,12 +152,14 @@
 				<List inset={isWideScreen.value || theme === 'ios'} strongIos>
 					<ListInput
 						type="text"
+						value={name ?? ''}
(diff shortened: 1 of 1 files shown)


guillemcordoba (approved):

guillemcordoba:
Agreed! Let's have a call where we set guidelines for our UI development, and we agree on best practices to use across the project.

Issue #143, opened 2026-03-06
[DESKTOP] Implement a way to upload a QR code image


guillemcordoba:
**Is your feature request related to a problem? Please describe.**
When someone shares a QR code image, there is no way to upload it from a Dash Chat instance running on desktop.

In desktop, the Add Contact page should have an "Upload" button next to the "Save" one that prompts you to select a qr code image.

PR #174, opened 2026-03-19
Refactor/break up AddContactPanel

jadehopepunk:
Before working on https://github.com/dash-chat/dash-chat/issues/143 I wanted to refactor AddContactPanel a bit. 

I've extracted out several components into the components/contacts directory.

My philosophy here is extracting only "pure" UI components that don't interface with context, they either display a thing, or do a thing and return some result. This makes it easier to add in the new upload button cleanly. All the layout stuff that positions them on a page should stay in AddContactPanel. Ideally that includes even margins around things and stuff.

One risk of this change is that I can't test the camera based scanning without having a real android device, or running it in a simulator, which I still haven't figured out. Maybe we could talk about this later, but also if you get a sec @guillemcordoba maybe you could regression test that for me? :)

And finally, while I think the components I extracted are not very controversial, I'd recommend going further and seperating out the visuals of AddContactPanel from the interface with the contacts store, decoding of the QR code, etc. This is partially done already, with helper functions, but I think there's a real joy in having pure components getting controlled by smart components. I haven't done that though, because if so I think we should discuss that as a refactoring direction. Possibly it's less of an ideomatically svelte approach, I'm not sure


ui/src/lib/components/layout/AddContactPanel.svelte
```diff
@@ -220,7 +215,7 @@
 								small
 								rounded
 								tonal={tab !== 'code'}
-								onClick={cancelScan}
+								onClick={() => void switchTab('code')}
```
guillemcordoba, line 218: Why this void?

guillemcordoba (approved):
Looks really good!

ui/src/lib/components/layout/AddContactPanel.svelte
jadehopepunk: Good point, missed that but will fix on the next branch

PR #178, opened 2026-04-15
Fix android development shell

guillemcordoba:
- Decoupled android and tauri devShells.
- Removed leftover ".dev" suffix artifact for the bundle id.
- Removed the preview on mobile.


jadehopepunk (approved):
Looks good

PR #180, opened 2026-04-15
Add weblate review worfklow

guillemcordoba:


jadehopepunk (approved):
I suggest we don't bother with human reviews for this kind of dev pipeline change. lgtm though :)

PR #181, opened 2026-04-15
Organize recipes into justfile modules

guillemcordoba:


jadehopepunk (approved):
Much neater, I like it

PR #192, opened 2026-04-23
Implement delete account

guillemcordoba:
Delete account's actual functionality was stubbed out, this implements it by deleting the data folder and restarting the app in desktop, and closing it in mobile (matching Signal's behavior). 

Fixes https://github.com/dash-chat/dash-chat/issues/209


crates/dashchat-node/src/node.rs
```diff
@@ -102,6 +102,14 @@ impl<R> CancelAndWait<R> {
         self.token.cancel();
         Some(self.handle.lock().await.take()?.await)
     }
+
+    /// Cancel the task and wait for it to finish, without consuming self.
+    pub async fn cancel(&self) {
```
jadehopepunk: This is pretty confusing. Based on the naming, it looks like the difference between  `cancel_and_wait` and `cancel` should be that only one of them waits. However, they actually both wait. 

It looks a lot like this new one also swallows the error.

Can't we end up with a single unified method that returns the error, and does not consume self?

crates/dashchat-node/src/node.rs
```diff
@@ -452,6 +460,16 @@ impl Node {
...
 
+    /// Stop background tasks and delete all account data.
+    pub async fn delete_account(&self) -> anyhow::Result<()> {
+        if let Some(ref task) = self.stream_task {
+            task.cancel().await;
+        }
+        self.mailboxes.clear().await;
+        std::fs::remove_dir_all(self.filesystem.data_path())?;
+        Ok(())
+    }
```
jadehopepunk: Given this is an important method, I wonder if it's worth having a clearly typed error. The error at the moment only occurs in `remove_dir_all`, but also I'd like to see if cancel can return an error, and is it possible for mailboxes clear to fail?

We don't want to leak detailed internal errors out to the client here I presume, so having this method log the error trace, and then return a more generic error "Failed to delete data path" might be good.

src-tauri/src/commands/account.rs
```diff
@@ -0,0 +1,26 @@
+use dashchat_node::Node;
+#[cfg(not(mobile))]
+use tauri::Manager;
+use tauri::{AppHandle, State};
+
+#[tauri::command]
+pub async fn delete_account(app: AppHandle, node: State<'_, Node>) -> Result<(), String> {
+    log::info!("Deleting account...");
+    node.delete_account().await.map_err(|e| {
+        log::error!("Failed to delete account: {e}");
```
jadehopepunk: I see the error is logged here. I'm not sure if this is better or worse that my previous suggestion of logging it in the node function. Either way, I think maybe this function should return a generic string error, and not use e.to_string() as that leaks internal error details to the front end. Am I thinking too much like a web developer there?

jadehopepunk (approved):
Looks like it will work. Some concerns around errors.

src-tauri/src/commands/account.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    // If we fail to shutdown, we still continue to delete all data and restart the app
+    match node.shutdown().await {
+        Ok(()) => {}
+        Err(ShutdownError::WaitOnDatabaseHandlesError(err)) => {
+            log::error!(
+                "Failed to wait on database handles closing, continuing to delete account: {err}"
+            );
+        }
+    }
```
maackle: Hmm, I wonder if we should tolerate the node *not* being able to shut down. Isn't it a bug on our end if we can't shut down cleanly?

I have a concern about the `PENDING_DELETE_MARKER` being an attack surface. If anyone can inject that file in the right place, they can bypass the user's input and trigger a deletion. It seems more secure to ensure that the deletion is done on the user's watch.

crates/dashchat-node/src/node.rs
```diff
@@ -570,10 +580,15 @@ impl Node {
...
-    pub async fn shutdown(mut self) {
-        if let Some(cancel_and_wait) = self.stream_task.take() {
-            cancel_and_wait.cancel_and_wait().await;
+    pub async fn shutdown(&self) -> Result<(), ShutdownError> {
+        self.mailboxes.clear().await;
+        if let Some(ref cancel_and_wait) = self.stream_task {
+            if let Some(Err(join_err)) = cancel_and_wait.cancel_and_wait().await {
+                return Err(ShutdownError::WaitOnDatabaseHandlesError(join_err));
+            }
         }
```
maackle, line 592: FYI unfortunately I had to totally change how shutdown works in the p2panda refactor, it's not a blocker for this PR, just info.

maackle (changes requested):
Would like to explore why we can't guarantee clean shutdown, it might just be an explanation for me as to why it's hard, but I want to ask

guillemcordoba:
@maackle I switched back to a proper shutdown method, but I had to change `OpStore` for that. Do you think we can merge this and tweak it if necessary when your changes are brought in?

ui/src/routes/settings/account/+page.svelte
```diff
@@ -22,16 +23,15 @@
 	async function handleDeleteAccount() {
 		loading = true;
 		try {
-			// TODO: Implement backend command for delete account
-			// await invoke('delete_account');
-			showToast(m.accountDeleted());
-			goto('/');
+			// On success the app restarts (desktop) or exits (mobile) immediately,
+			// so no code after this line executes on the happy path.
```
maackle:
```suggestion
			// On success the app exits immediately,
			// so no code after this line executes on the happy path.
```

maackle (approved):

PR #193, opened 2026-04-24
Push notifications retrying is unified and made robust

guillemcordoba:
- Updated tauri-plugin-notification to its latest version
- Added SingletonTaskWithRetries to dashchat-utils: a self-spawning task with exponential backoff that ignores duplicate triggers while running, plus 8 unit tests
- Refactored push notifications to use SingletonTaskWithRetries for FCM register/unregister, eliminating competing retry loops when the user toggles notifications
- Simplified set_notifications_enabled to save the setting first and trigger the watcher, instead of blocking on infinite retries


jadehopepunk (approved):
That looks great. No notes. :shipit:

PR #194, opened 2026-04-24
Test banner

jadehopepunk:
Add a temporary banner indicating to users that this is test software only


guillemcordoba:
Looks good, but I'd request two changes:

1) The banner leaves no margin for the top buttons in iOS: 

IMG_0187

2) The "EARLY ACCESS TEST" span is not vertically aligned to the center, to match the "FEEDBACK" button.

PR #207, opened 2026-04-27
Restyle site, setup redirects to App Store and Play Store

guillemcordoba:


maackle (approved):
I haven't reviewed the actual visual effects, and I see no problems with the code, so this is kind of a non-review

PR #212, opened 2026-04-28
Fix/208 scroll up when keyboard displays

jadehopepunk:


guillemcordoba:
Tested on iOS, it works well.

PR #214, opened 2026-04-28
Proper iOS push notifications support

guillemcordoba:
Depends on https://github.com/dash-chat/dash-chat/pull/215

Migrated local stores from redb to SQLite with support for multiple readers and writer processes so the iOS Notification Service Extension and main app can share the same folder in the filesystem.

Added per-op `author: PublicKey` to NotifyTopicsRequest so the server filters authors out of subscribers — iOS can't programmatically suppress a delivered notification without Apple's usernotifications.filtering entitlement, so we avoid sending the push entirely.                                                                                                                                           

Wired thread-id through the APNS payload and plugin so notification.group is set on iOS, fixing tap-to-navigate.

This breaks backwards compabitility. A follow up PR will address storing the node files in a folder dependant on the app version.

Another follow up PR will fix the issue of notifications being shown even for active chats.


crates/dashchat-node/src/topic.rs
```diff
@@ -137,6 +138,28 @@ impl p2panda_spaces::traits::SpaceId for TopicId {}
 
 pub type DashChatTopicId = TopicId;
 
+// -- SQLite encoding for TopicId --
```
jadehopepunk, line 141: Maybe a note for later.  I'm not sure if we should spread the sqlx code all through things. It might be nice to have a clear DB layer to make it easier to switch datastores. I don't suggest changing that in this PR.

jadehopepunk (commented):
LGTM

I'm not sure I'd be able to spot problems with the Swift.

Also, some of Claude's suggestions seem important, particularly the "JSON string interpolation in NotificationService.swift (line 59)"

jadehopepunk (approved):
Meant to approve, sorry

PR #223, opened 2026-04-29
Work around plugin requestPermissions() bug

guillemcordoba:
https://github.com/dash-chat/dash-chat/pull/220 didn't fix the bug (it's still good to have it merged since we want the additional 3MB downloaded at app download time instead of the first time a QR code is scanned).

This PR works around a bug in the plugin that made it so that the `await requestPermissions()` never returned.


jadehopepunk (approved):
A pain we need to do that of course, but it seems fine. It's good that it's hidden in the util file there.

PR #227, opened 2026-04-29
Refactor: use flex's `column-reverse` in the message list layout

### LINKED ISSUE #206: [MOBILE] After sending a message, the message input is over the latest messages
Image

**Describe the bug**
The text box, when there isn't a keyboard, covers the bottom messages and obscures them. There should by default be space between the new text box and the last text. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

guillemcordoba:
Fixes https://github.com/dash-chat/dash-chat/issues/206. 

This PR fixes a class of problems around keeping the messages list scrolled to always show the latest messages. Having a `column-reverse` flex direction allows us to move away from an imperative style of keeping the scroll to the bottom always, to a declarative style where we don't need to do anything once a new message appears, it's already scrolled to the bottom of the page.


ui/src/lib/components/messages/ScrollToBottomButton.svelte
```diff
@@ -0,0 +1,44 @@
...
+	interface Props {
+		unreadCount?: number;
+		onClick: () => void;
+	}
+
+	let { unreadCount = 0, onClick }: Props = $props();
+</script>
+
+<button
+	class="relative flex h-10 w-10 items-center justify-center rounded-full bg-gray-100 shadow-md transition-opacity hover:bg-gray-300 dark:bg-gray-700 dark:hover:bg-gray-600"
```
jadehopepunk, line 17: God I hate tailwind. :) This sort of line just has to re-make so many decisions about our semantic button choices.

ui/src/lib/components/ReverseScrollPage.svelte
```diff
@@ -0,0 +1,344 @@
+<!--
```
jadehopepunk, line 1: That's a lot of comment. I think we should chat about how we're going to use comments. I can see that the more we get claude involved in this, the easier it is to write comments, and maybe there's good use for them if they're read by claude. 

We know that it can be hard to keep them in sync with code, and maybe our LLM code reviews will help with that. In this case however, I worry that it's such a long comment, and everything in it will get treated as gospel by claude.

ui/src/lib/components/ReverseScrollPage.svelte
```diff
@@ -0,0 +1,344 @@
...
+	</div>
+</Page>
+
+<style>
+	/*
+		Konsta's `.k-page` is `absolute overflow-auto` by default. We use it
+		as a positioning context for our absolute scroll element + sibling
+		navbar, but we don't want it to scroll itself.
+	*/
+	:global(.reverse-scroll-host) {
```
jadehopepunk, line 324: I like the abstract nature of this component.

ui/tests/pages/direct-chat.ts
```diff
@@ -82,6 +84,86 @@ export function messagesContainer() {
 	return document.querySelector(selectors.messages);
 }
 
+/** True if the chat scroll container is pinned to the bottom (column-reverse). */
+export function isScrollAtBottom(): boolean {
```
jadehopepunk, line 88: We have these cool components like ReverseScrollPage, but then we have functions like this that calculate things about it from first principles. Maybe it's fine, because we're just looking at the scroll selector. It'd be nice if we can ask "ReverseScrollPage" this question somehow though. I reallise you don't really ask components questions. They can pass things to their children though, like "isAtBottom", or they can have callbacks. Not sure if any of those things would be valuable approaches.

ui/src/lib/utils/chat.ts
```diff
@@ -0,0 +1,7 @@
+/** Distance from the bottom (in px) below which we consider the user
+ * "at the bottom" of the chat — controls when the scroll-to-bottom
+ * button hides and when self-sends snap back to the bottom.
+ *
+ * Shared between the chat page and its E2E test helpers so both stay
+ * in sync if the value is tuned. */
+export const SCROLL_BOTTOM_THRESHOLD = 200;
```
jadehopepunk: I don't love this. It seems like this is an internal constant of ReverseScrollPage. It's not some general app setting.

justfile
```diff
@@ -19,3 +19,8 @@ mod ios 'scripts/ios.just'
 # cut a new release (e.g. just release 0.11.0)
 release version:
     ./scripts/release.sh {{version}}
+
+# format both UI and rust files
+format:
```
jadehopepunk, line 29: 👍

jadehopepunk (approved):
I left comments, but overall I like this change a lot. Seems like a good approach. Some great generic bits of code that pull the complexity of it out of the direct-chat page. I've given it a quite test, but not actually on mobile.

ui/tests/pages/direct-chat.ts
guillemcordoba: Hum this function is a test helper that's supposed to be called at e2e test time, when the app has been compiled. Instead of manually calling the UI elements with `querySelector()` and inspecting the scroll, it's nice to be able to call these functions.

At that the moment of the e2e tests, svelte components don't exist as such, this function can't interact with `ReverseScrollPage` because it's been compiled down to the JS bundle. 

Am I missing something?

@jadehopepunk

PR #236, opened 2026-05-01
Extract tracking of read messages into a reusable action

guillemcordoba:


jadehopepunk (approved):
Nice one, I really like this

PR #237, opened 2026-05-01
Refactor: the direct chats layout owns the subscriptions for the page

### LINKED ISSUE #222: [ANDROID] Going from the direct messages settings view to the direct messages view glitches while loading
**Describe the bug**

When going back to the direct messages page from the direct messages settings page, the direct messages page shows no messages and a white background navigation bar until the messages are loaded, and then the page looks fine.

**To Reproduce**
1. Go to the direct messages settings.
2. Go back.

**Expected behavior**

The page has a consistent loading behavior that doesn't glitch.

**Screenshots**

 - Device: TCL 40
 - OS: Android 14
 - Version: v0.13.0

**Additional context**

guillemcordoba:
Fixes https://github.com/dash-chat/dash-chat/issues/222


jadehopepunk (changes requested):
I can see the explanation for why this is done in the comments, and having the flash of a blank screen sounds like an annoying thing that we don't want, but this clearly makes the code worse.

A layout should be a layout. A page should load the data. The fact that the layout happens to stay mounted during the page switching in question also seems like a bit of an accident anyway.

Another approach, which I don't love, could be to have the page cache the relevant data in local component state and only change it with an effect (if the input props change). Even then, in react, I'd probably wrap the component in another component that does the caching, so we can have a direct chats layout and a cached direct chats layout.

jadehopepunk (approved):
Approved after discussion, but maybe move the contacts store up to a higher level layout.

jadehopepunk (approved):
I like this much better for sure. It moves the complexity out of the layouts.

I'm not totally clear on what it caches. I would have thought you'd need to specify which stuff we're caching within the layout. Seems good though.

guillemcordoba:
It stores a subscription to each signal that any of the children declare, and unsubscribes from it when the element is destroyed. I agree that caching is not the right name, renaming to `KeepAliveScope`.

PR #238, opened 2026-05-01
Fix frontend warnings and make them fail CI from now on

jadehopepunk:


ui/src/lib/components/profiles/ProfileAvatar.svelte
```diff
@@ -9,7 +9,9 @@
 
 	const contactsStore: ContactsStore = getContext('contacts-store');
 
-	const profile = useReactivePromise(contactsStore.profiles, chatActorId);
+	const profile = $derived(
```
guillemcordoba, line 12: Why was this necessary? If this is necessary there are a lot of places in the code we need to change.

guillemcordoba (approved):
Looks great! Thanks for doing this.

ui/src/lib/components/profiles/ProfileAvatar.svelte
jadehopepunk: Yes, it's necessary because chatActorId is an incoming prop. If we don't use derived, we get the warning: 

_"This reference only captures the initial value of `chatActorId`. Did you mean to reference it inside a closure instead?"_

PR #239, opened 2026-05-04
Fix first chat tooltip position

### LINKED ISSUE #225: [IOS] "Start your first chat" notice's position is broken after the test banner.
**Describe the bug**

"Start your first chat" notice's position is broken after the test banner.

**To Reproduce**
1. Install the app in iOS.
2. Create a profile.
3. See the position of the "Start your first chat here tooltip".

 - Device: iPhone
 - Version: v0.13.0

guillemcordoba:
Fixes https://github.com/dash-chat/dash-chat/issues/225


ui/src/routes/+page.svelte
```diff
@@ -43,6 +48,15 @@
...
 				</Link>
+				{#if !isWideScreen.value}
+					{#await $chatSummaries then chats}
+						{#if chats.length === 0}
+							<div class="absolute end-0 top-full mt-2 z-30">
+								<FirstChatTooltip />
+							</div>
+						{/if}
+					{/await}
+				{/if}
```
jadehopepunk, line 56: Not totally sure why rightClass has relative and then this absolute div goes in the left. Also, why is it absolute? Stuff always goes wrong with absolute.

jadehopepunk (approved):
Despite the comment, certainly seems safe to ship

ui/src/routes/+page.svelte
guillemcordoba: To anchor the first chat tooltip to the button, in a way that if that button moves, the tooltip moves with it. With the previous pattern, after adding the test banner, the position of the tooltip was wrong. This makes the position of both coupled, so that when we remove the test banner, the position of the tooltip is still correct.

Issue #235, opened 2026-04-30
[IOS] Notifications are displayed for active chats


guillemcordoba:
**Describe the bug**

On iOS, when you receive a message, that message always shows a push notification, regardless of whether the message that was received is part of the chat that the user has opened in the app. 

**To Reproduce**
1. Go to the direct messages page.
2. Send a message to the user that is using the iPhone for the direct messages chat that is open in the iPhone.
4. See the notification appear in the system notification tray.
**Expected behavior**

If the user already has the page for a chat open, no notifications should be shown for messages that are part of that chat.

**Device (please complete the following information):**
 - Device: iPhone 12
 - Version: 0.13.0

PR #242, opened 2026-05-04
Don't show notifications for active route on iOS

guillemcordoba:
Fixes https://github.com/dash-chat/dash-chat/issues/235. 

Tested in both Android and iOS.


guillemcordoba:
Tested all cases that claude is warning about, I confirm they work correctly.

maackle (approved):

PR #247, opened 2026-05-05
Backward compatibility types

maackle:
Add types for backward compatibility

- New crate comcap with "common" types and patterns, which may not be useful now but may be in the future
- Placeholder for a future version 1 of chat message with media, backward compatible with medialess "version 0" messages
- unversioned types from older clients deserialize seamlessly into version 0. Adding versioning to an unversioned type is backward compatible

Ironically, this is failing the backward compat test in CI, however the followup PR #248 is passing, so let's let that be the guide. @guillemcordoba also, it's hard to tell *why* the CI tests are failing, would be good if errors were logged.


crates/dashchat-node/src/chat/message.rs
```diff
@@ -13,11 +17,113 @@ use serde::{Deserialize, Serialize};
...
+impl PartialOrd for ChatMessageContent {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        (self.message(), self.media()).partial_cmp(&(other.message(), other.media()))
+    }
+}
+
+impl VersionConvert for ChatMessageContent {
+    type Capabilities = Capabilities;
+
+    // TODO: just take Capabilities?
```
maackle, line 100:
```suggestion
```

jadehopepunk (approved):
LGTM ;)

PR #248, opened 2026-05-06
Compat integration

maackle:
Integrates compatibility checking for messages:

- [x] stores compatibility of each contact
- [x] determines "highest common compatibility" (infimum) of all chat members
- [x] downgrade all new messages to infimum


crates/dashchat-node/src/contact.rs
```diff
@@ -36,6 +36,14 @@ pub struct QrCode {
     pub inbox_topic: Option<InboxTopic>,
     /// The intent of the QR code: whether to add this node as a contact or a device.
     pub share_intent: ShareIntent,
+
+    /// Capabilities supported by this node, for version negotiation.
+    #[serde(default = "default_capabilities")]
+    pub capabilities: Capabilities,
```
guillemcordoba: What if the user updates their app after they generate their contact code?

guillemcordoba (commented):
Looking good! I want @jadehopepunk to also have a look at this before merging. I think my main blocker is that it seems to me like we need to take into account app updates after the contact code has been generated. There are different ways to implement it, one could be an extension in all operations that describes the data schema version that the author is at, another could be a single "DataSchemaUpgrade" operation that they commit into their announcements topic. Also taking into account that capabilities are per device, and not per agent.

crates/dashchat-node/src/stores/local_store.rs
```diff
@@ -258,6 +281,42 @@ impl LocalStore {
...
         Ok(())
     }
+
+    pub async fn get_device_capabilities(
+        &self,
+        device_ids: impl IntoIterator<Item = DeviceId>,
+    ) -> anyhow::Result<Option<Capabilities>> {
+        Ok(
+            QueryBuilder::new("SELECT (messaging) FROM capability_versions WHERE device_id IN ")
+                .push_tuples(device_ids, |mut qb, device_id| {
```
jadehopepunk: I hate how sqlx doesn't have a cleaner way to do `IN` statements.

jadehopepunk (approved):
Looks like a really good approach

maackle:
Well @guillemcordoba @jadehopepunk I completely re-did this. I realized I don't need a database table to store capabilities, that capability info should come directly from the Announcements topic. Thanks Guillem for pointing this out. My first PR on this #177 actually did that, but I forgot to re-do it for this PR. I think this is a better solution, and it's well tested with version upgrades in chats now too

EDIT: I am storing capabilities in the database anyway since it's useful caching, and also also took this opportunity to store profile in the db as well

guillemcordoba (approved):
I'm unconvinced about the need for caching, but I'm happy adding it if we don't hit any inconsistencies with the state duplication. Thanks for all the work!

PR #266, opened 2026-05-08
Keep alive all chats summaries

### LINKED ISSUE #202: Contact list disappears when coming back from a conversation
**Describe the bug**
I can make my list of contacts disappear

**To Reproduce**
1. Have some contacts
2. Go to a chat 
3. Go back
4. See a blank list of contacts

**Expected behavior**
The contact list should not disappear

**Screenshots**

 - Device: Pixel 9a
 - OS: Android
 - Version 0.12.1

**Additional context**

guillemcordoba:
Fixes https://github.com/dash-chat/dash-chat/issues/202

### DIFF
--- a/ui/src/lib/stores/KeepAliveScope.svelte
+++ b/ui/src/lib/stores/KeepAliveScope.svelte
@@ -6,7 +6,8 @@
 	import {
 		KEEP_ALIVE_SCOPE_KEY,
 		type KeepAliveScope,
-	} from './keep-alive-scope.svelte';
+		trackRpVersion,
+	} from './keep-alive.svelte';
 
 	let { children }: { children: Snippet } = $props();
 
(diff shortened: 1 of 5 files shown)


maackle (commented):
I don't understand the problem well enough to know if this is the right fix. Do you know why the contact list was showing up blank? This seems to be caching its previous value, so that it doesn't need to recompute itself, but why, when it recomputes itself, does it show up blank, instead of showing the contact list, maybe with some delay?

guillemcordoba:
After investigation, this was an internal bug in the previous version of signalium. Signalium v2 tore down reactive() state eagerly on unwatch and swapped pending promises via  _setPromise, so any await rp in flight at the moment a subscriber dropped got orphaned  forever - which is what made chat2's summary() chain hang and the contact list stay blank on home re-mount.

Closing in favor of https://github.com/dash-chat/dash-chat/pull/269

PR #267, opened 2026-05-11
Refine e2e tests

guillemcordoba:
- Logs for each test agents directed to stdout.
- Refactor so that test utils are directly callable from the e2e tests.
- Extracted scroll behavior test from the full flow test.
- Extract tauri build command from the WebDriver config.
- Fixed typescript diagnostics for the e2e tests package.


maackle (approved):
Good improvements

PR #272, opened 2026-05-12
Workaround missing entitlement

guillemcordoba:
- Workaround missing entitlement by re-registering mDNS for local mailbox server.
- Stable mailbox ID derived from the `DeviceId`.

This fixes an issue with the current code in this scenario:

0. Dash Chat is installed in a Mac and in and iPhone that are connected to the same network.
1. The user opens Dash Chat in Mac, and runs the Local Mailbox Server (while Dash Chat in the iPhone is not opened!). 
2. The user opens Dash Chat in the iPhone. 
3. The user tries to send messages from the Mac app to the iPhone app and viceversa.

The root cause of the bug is the missing [multicast entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.developer.networking.multicast) for the iOS app. Unfortunately, this entitlement can only be added after requesting it to Apple, and have them review and accept the request, which could take some weeks. For now, I suggest merging this as a workaround, and when we receive the entitlement, we can easily remove it.


maackle:
Hmm, according to deepwiki, this shouldn't be necessary: https://deepwiki.com/search/does-a-registered-service-rean_3c9b940a-51e7-49e1-a661-15279d3bf01f?mode=fast (I learned a lot here that I didn't know)

- Clients who are browsing a service type will immediately query the network for existing records, so that should cover a new phone joining before the LMS registered itself
- Service registration never expires. The purpose is to respond to client queries. TTL only applies to client-held data and dictates when the client will make a fresh query

So according to the deepwiki answer, in your scenario step 3, the phone should query the mDNS service when it calls `.browse()`, and learn about all LMSes that way. Maybe the deepwiki is wrong, or maybe there's some other problem?

maackle (approved):
I can get behind this, but I almost wonder at this point if we just want to say LMS are not supported on iOS until the entitlement comes through. Especially if we don't need to demo LMS on iOS in the next few months

guillemcordoba:
We do need to demo LMS with iOS in the next few weeks. I'm going to merge this as is, and when the entitlement comes through, I'll create a PR to remove it.

PR #278, opened 2026-05-13
Group android notifications by chat

### LINKED ISSUE #274: [ANDROID] Push notifications for the same chat are not grouped together in the notifications system tray
**Describe the bug**
Push notifications for the same chat are not grouped together in the notifications system tray

**To Reproduce**
1. Send multiple messages in the same chat to an Android device with push notifications enabled.
2. Inspect the system notifications tray.

**Expected behavior**

Notifications for the same chat should be grouped together in the system tray.

**Screenshots**

 - Device: TCL 40
 - OS: Android 14
 - Version: v0.15.0

guillemcordoba:
- New plugin version
- Use MessagingStyle

Fixes https://github.com/dash-chat/dash-chat/issues/274

### DIFF
--- a/src-tauri/src/push_notifications/receive_push_notification.rs
+++ b/src-tauri/src/push_notifications/receive_push_notification.rs
@@ -184,41 +184,62 @@ async fn handle_push_notification(
                 }
             };
 
-            let sender_name = if let Some(agent_id) = sender_agent_id {
-                node.local_store
-                    .get_profile(agent_id)
-                    .await
-                    .ok()
-                    .flatten()
-                    .map(|profile| profile.name)
-            } else {
-                None
+            let Some(sender_agent_id) = sender_agent_id else {
+                return Ok(Some(NotificationData {
+                    title: Some(sonix_i18n::t!("newMessage")),
+                    icon: Some("ic_stat_icon".to_string()),
+                    ..Default::default()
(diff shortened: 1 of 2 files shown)


src-tauri/src/push_notifications/receive_push_notification.rs
```diff
@@ -197,28 +197,38 @@ async fn handle_push_notification(
...
                 route: Some(chat_route),
                 ..Default::default()
-            }))
+            };
+
+            // Android-only: reusing a stable id replaces notifications on iOS.
+            #[cfg(target_os = "android")]
+            if direct_chat_agent_id.is_some() {
+                notification_data.id =
+                    i32::from_le_bytes([topic_id[0], topic_id[1], topic_id[2], topic_id[3]]);
```
maackle, line 227: This is reasonable. I notice this kind of truncation happening elsewhere too, and it's good practice when truncating IDs down to a small number of bits to also XOR those bits with something context-specific so people aren't able to mine their topic IDs to cause collisions in the truncated form. In this case XORing with a node-specific secret would be good. Not sure if we have something like that handy other than the private key, which would seem risky, but it could be worth storing a node-specific random salt in the db just for this kind of purpose.

maackle (approved):

PR #288, opened 2026-05-18
Track mailbox sync states, expose sync and connection states to the UI

guillemcordoba:
- Rename MailboxTracker to MailboxConnectionState with last_success_at/last_error.
- Tracks per-mailbox sync state via new SQLite-backed MailboxSyncTracker.
- Exposes sync and connection state to UI via four Tauri channel commands and a signalium-backed MailboxTrackerStore.

This is necessary to enable the UI to show when the app is connected to the cloud mailbox server or to how many local mailbox servers, or to render which messages have been synced with which mailboxes.


crates/mailbox-client/src/manager.rs
```diff
@@ -84,24 +111,54 @@ impl MailboxTracker {
...
-        }
+impl<Item: MailboxItem> TrackedMailbox<Item> {
+    fn new(client: Arc<dyn MailboxClient<Item>>) -> Self {
+        let (tracker, _) = watch::channel(MailboxTracker::new());
+        Self { client, tracker }
+    }
+
+    pub fn tracker(&self) -> watch::Receiver<MailboxTracker> {
+        self.tracker.subscribe()
+    }
```
maackle: strange name

crates/mailbox-client/src/manager.rs
```diff
@@ -255,84 +344,89 @@ where
...
 
     /// Immediately sync the given topics with the given mailbox:
     /// - Ensure all items held by the mailbox are fetched
     /// - Publish any items that the mailbox is missing to the mailbox
-    pub async fn sync_topics(
+    async fn sync_topics(
         &self,
         topics: impl Iterator<Item = Item::Topic>,
-        mailbox: Arc<dyn MailboxClient<Item>>,
+        mailbox_id: &MailboxId,
```
maackle: id() is already on MailboxClient

maackle (commented):
I can read more thoroughly if you include a high level description of what this does

maackle (approved):
Seems like a shame how much plumbing there needs to be for this, but maybe there's no way around it. Good work!

PR #289, opened 2026-05-19
Extract message related elements from the direct chats page

guillemcordoba:


jadehopepunk (commented):
This is great, so happy to see this in here. 

You might want to consider Claude's nit #1 regarding dead imports.

jadehopepunk (approved):

PR #291, opened 2026-05-20
Transition UX for disconnected, local mailboxes, and cloud mailbox states

guillemcordoba:
- Add message sent indicator showing per-mailbox sync status visually
- Show connection state transition UX for cloud and local mailboxes
- Cover new indicators and connection states with e2e tests


crates/mailbox-client/src/manager.rs
```diff
@@ -224,19 +235,24 @@ where
         // TODO: check for existing mailbox with different ID but same "URL" (which is currently abstracted away and inaccessible here, darn)
         // TODO: make the ID come from the mailbox server itself, e.g. for mDNS discovery the ID is set by the mDNS service, but multiple services could point to the same actual mailbox state.
```
maackle, line 236: this would still be good to do at some point, so that the ID is backed by a signature and can't be hijacked by just anyone

maackle (approved):
I mainly reviewed the rust parts, but they look good

jadehopepunk (approved):
This looks really good. Nice work on the e2e tests in particular

PR #294, opened 2026-05-21
Create new group

jadehopepunk:
Relates to the [Create new group](https://miro.com/app/board/uXjVJgzq4Mo=/?moveToWidget=3458764673179323704&cot=14) user story.


crates/dashchat-node/src/node/stream_processing.rs
```diff
@@ -323,8 +323,10 @@ impl Node {
         let topic = header.extensions.topic;
 
         match &payload {
-            Payload::Chat(ChatPayload::JoinGroup { .. }) => {
-                // Nothing to do.
+            Payload::Chat(ChatPayload::JoinGroup { chat_id }) => {
+                if let Err(err) = self.join_group(*chat_id).await {
```
maackle, line 327: huh, I guess I relied on this getting called by test behavior. Sorry if this was hard to find.

maackle (approved):
Looks good other than Claude's recommendations which I mostly agree with (no need to localize the temporary "not implemented" strings though)

PR #303, opened 2026-05-26
Pin node to a specific version, and set it to node 22

jadehopepunk:
We previously used node 20 in some github workflows, and didn't specify it during development at all.


maackle (approved):
Nice, I can't speak to the nix stuff but I approve of this locking-down!

PR #309, opened 2026-06-02
Wire up sending group messages 

guillemcordoba:
Sending group messages for a group you have created works!

I know the group chats store has some code duplication with the direct chats store, but I thought it was best to duplicate this contained logic until we have implemented all the user stories from group chats, and then refactor to extract the pieces of functionality that are shared.

This PR intentionally leaves out any e2e tests, because for them to make sense members of created groups should see them in the chats list, but they don't see them yet.


jadehopepunk (approved):
That looks great. I see your point about e2e tests and how we'll have to add those later, that makes sense.

Fine on the duplication for now. 

Clause is still winging about multi-device confusion, obviously we should ignore that.

I agree with claude that there's maybe some room to make sendMessage handle failures more robustly, but later for that seems fine.

PR #314, opened 2026-06-03
Fix group member profile in group chat page

guillemcordoba:
- Return a new `GroupMember` struct from the `get_group_members()` command. 
- Renamed `GroupMemberData` to `GroupMember` in the frontend.
- Renamed `GroupMember` to `GroupMemberWithProfile` in the frontend.
- Show the profile for the group member with the author's public key in the array of devices_ids.


jadehopepunk (approved):
Looks good. There's a format issue stopping from merging right now, if I get a sec today I'll fix.

PR #307, opened 2026-05-29
Debugging p2panda node integration

maackle:
This is based off of https://github.com/sandreae/dash-chat/tree/p2panda-node-integration

- incorporates `aliased` crate for better debug experience
- fixes "mark op processed" -- group ops as well as app ops need to be marked for `consistency` to work

PR #295, opened 2026-05-21
p2panda node integration

sandreae:
Integration of [p2panda node](https://crates.io/crates/p2panda) into dashchat-node.

## Node actor

I introduced a dedicated actor to handle all p2panda related tasks like publishing, subscribing, importing (mailbox stream) etc... it is a very thin wrapper around the existing node API with the addition that it holds all current publishing handles and subscriptions and combines operations arriving on all topics into one stream. It is also where groups control messages are processed. When publishing control messages locally I added a wrapper around `PublishFuture` which only completes after after all application processing has finished. Events are forwarded further onto the application processor, in the case of groups or application operations they come with a `processed_tx` oneshot sender for signalling that all processing is complete, this will allow the `ProcessorFuture` to complete. In the case of groups operations, then an optional error is also included in the event.

## Type changes

Due to type changes (p2panda node uses it's own `LogId` and `Topic` types) some changes were required here and I introduced conversions between the dashchat `Topic` where required. Validation for topic type correctness occurs.

## Registering bootstrap nodes

This is an area which needs follow-up work. After initial conversations with the dashchat team I understand that mailboxes will provide a discover mechanism by which they inform connected users who their potential bootstrap nodes are. I have taken an extremely naive approach in this PR (look into operations we receive from the mailbox and add every author as a bootstrap) to demonstrate that the approach is viable, however I recommend this is immediately replaced with something more well thought-through. Things to consider are the liveness of known peers and making sure common relays are used. I can imagine that a dedicated mailbox API should be used to get bootstrap node information.

## Node configuration

I added two new arguments to the node configuration struct, these are `mdns_mode` for configuring behaviour around mdns discover, and `relay_url` for configuring an iroh relay. By default `mdns_mode = true` and relay url uses one of iroh's publicly available servers. There are many more node configuration options which you may with to expose.

## Tests

I introduced basic unit tests for the node actor and some additional integration tests for the new p2p functionality. Some existing tests are now failing but I'm not sure if it's because of what I've introduced or existing an existing issue. I ran two instances of the tauri app locally and successfully established a direct chat and sent messages without a mailbox server running :partying_face:  

I am seeing some flaky tests when running all tests in dashchat-node via `cargo nextest run`, some fail when all are running but not when run individually. If this issue occurs because of my changes I can look into it more.

## p2panda version

[...]

PR #320, opened 2026-06-03
p2panda-node integration, enabling p2p sync

maackle:
This is a merging of https://github.com/dash-chat/dash-chat/pull/295 and https://github.com/dash-chat/dash-chat/pull/307, rebased onto develop

This updates to a new p2panda version with significant refactors and name changes, most notably `PublicKey` -> `VerifyingKey`, and includes a ton of work by @sandreae, thank you!


crates/dashchat-compat/Cargo.toml
```diff
@@ -4,8 +4,8 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
+aliased = { workspace = true }
```
maackle, line 7: this is the new and improved replacement for named-id which allows us to get off the p2panda fork

maackle:
@sandreae so `group_chat_capability_upgrade` is the most commonly failing test now, due to timeouts.

I notice that it consistently runs into this warning:

```
WARN dashchat_node::node::actor: crates/dashchat-node/src/node/actor.rs:286: groups processing error: Groups("duplicate operation e2db8ecb2760829bb0623497ca958ea90ddb42990a7a8fc34698042699d5d405 processed in group 77ce170b51be18b32296e73c862f89c56d868ee201b952a3c0860d1b5f26fcb1")
```

here: https://github.com/dash-chat/dash-chat/blob/89c9e7f799681686075623a5d75e32a66af24105/crates/dashchat-node/src/node/actor.rs#L286

Is this expected/acceptable and could this have any effect on the test?

maackle:
@guillemcordoba is it ok for these SQL tables to be regenerated? (public_key field becomes verifying_key) https://github.com/dash-chat/dash-chat/blob/89c9e7f799681686075623a5d75e32a66af24105/crates/push-notifications-server/src/driver/sql.rs#L30-L56

sandreae:
> @sandreae so `group_chat_capability_upgrade` is the most commonly failing test now, due to timeouts.
> 
> I notice that it consistently runs into this warning:
> 
> ```
> WARN dashchat_node::node::actor: crates/dashchat-node/src/node/actor.rs:286: groups processing error: Groups("duplicate operation e2db8ecb2760829bb0623497ca958ea90ddb42990a7a8fc34698042699d5d405 processed in group 77ce170b51be18b32296e73c862f89c56d868ee201b952a3c0860d1b5f26fcb1")
> ```
> 
> here:
> 
> https://github.com/dash-chat/dash-chat/blob/89c9e7f799681686075623a5d75e32a66af24105/crates/dashchat-node/src/node/actor.rs#L286
> 
> Is this expected/acceptable and could this have any effect on the test?

This (operations sometimes arriving more than once) is expected and it shouldn't cause any issues. The error is a bit misleading, they should rather just be ignored (maybe with a trace log).

.config/nextest.toml
```diff
@@ -0,0 +1,177 @@
...
+# subsequent attempt to be marked as flaky. Can be overridden through the
+# `--retries` option.
+# Examples:
+# * retries = 3
+# * retries = { backoff = "fixed", count = 2, delay = "1s" }
+# * retries = { backoff = "exponential", count = 4, delay = "1s", jitter = true, max-delay = "10s" }
+retries = 2
+
+# Whether to treat flaky tests as passing or failing.
+flaky-result = "pass"
```
guillemcordoba, line 13: Ideally now that we don't have pressure to release this we shouldn't treat flaky tests as passing.

crates/dashchat-node/src/node/actor.rs
```diff
@@ -0,0 +1,625 @@
...
+use tokio_stream::{StreamExt, StreamMap};
+use tracing::warn;
+
+use crate::Payload;
+use crate::stores::GROUPS_STATE_ID;
+
+type GroupsProcessor = p2panda_auth::processor::GroupsProcessor<Topic, Extensions, LogId>;
+
+/// Node actor commands.
+pub(crate) enum Command {
```
guillemcordoba, line 28: I think we should refactor this away in the near future. For now it's fine to merge as is.

crates/dashchat-node/src/node/app_processing.rs
```diff
@@ -0,0 +1,414 @@
...
+        // @TODO: this requires a reliable way to know the agent id from the device id
+        // even if they're not a contact.
+        let known = self
+            .local_store
+            .lookup_contacts(member_device_ids.iter())
+            .await?;
+
+        for agent_id in known.into_values() {
+            let topic = Topic::announcements(agent_id);
+            self.initialize_topic(*topic).await.map_err(|err| {
```
guillemcordoba, line 263: @maackle does this superseed the functionality from https://github.com/dash-chat/dash-chat/pull/329?

crates/dashchat-node/src/node/app_processing.rs
```diff
@@ -0,0 +1,414 @@
...
+
+        Ok(())
+    }
+
+    async fn register_bootstrap(
+        &self,
+        operation: &ProcessedOperation<Payload>,
+        source: &Source,
+    ) -> anyhow::Result<()> {
+        // Dash Chat relies on mailbox servers for discovering bootstrap
```
guillemcordoba, line 430: I agree with this comment, and I think our best approach here would be to add a p2panda bootstrap endpoint to our mailbox servers. This is fine to merge for now.

crates/dashchat-node/tests/bootstrap.rs
```diff
@@ -0,0 +1,51 @@
...
+        .add_mailbox_client(mailbox.client())
+        .await;
+
+    alice
+        .behavior()
+        .initiate_and_establish_contact(&bobbi, ShareIntent::AddContact)
+        .await
+        .unwrap();
+
+    // Drop the mailbox meaning it can no-longer provide sync for alice and bobbi. If they
```
guillemcordoba, line 37: Shouldn't this test expect bob to receive the message to guarantee it's actually syncing?

crates/dashchat-node/tests/compat_upgrades.rs
```diff
@@ -243,6 +260,11 @@ async fn group_chat_capability_upgrade() {
         .add_mailbox_client(mailbox.client())
         .await;
 
+    println!(
```
guillemcordoba, line 240: Do we want to be println!()ing in tests?

src-tauri/src/push_notifications/receive_push_notification.rs
```diff
@@ -0,0 +1,331 @@
+use std::path::PathBuf;
```
guillemcordoba: I think this file should not exist, it's in `src-tauri/src/notifications/push_notifications/receive_push_notifications.rs` in develop.

guillemcordoba (approved):

.config/nextest.toml
maackle: Ideally yes but in reality they are still flaky, so let's look at that as a follow up

crates/dashchat-node/src/node/actor.rs
maackle: You mean refactor the actor pattern into just methods? I would like that too but it's significant

crates/dashchat-node/src/node/app_processing.rs
maackle: I think this is more to do with the fact that we may get group operations out of order, so we just do the extreme overkill of subscribing to every member on every new message.

crates/dashchat-node/tests/bootstrap.rs
maackle: Sounds reasonable, I added the check

crates/dashchat-node/tests/compat_upgrades.rs
maackle: I think it's helpful. lots of tests have debugging enabled. nextest only shows output on failure

maackle:
@guillemcordoba i wonder if you can look into the (as of last merge) 2 CI checks that are failing, one being e2e tests, one mailbox-server nix stuff

crates/dashchat-node/src/node/app_processing.rs
sandreae: Two features will come in with the spaces integration which will be useful here:
1) spaces operations (currently groups operations) be emitted from the node in their causally sorted order
2) they will come on events containing spaces metadata, including things like `current_membership`, `added_members`, `removed_members` etc...

Especially with `2)` you then won't have to make a call to the database to retrieve this information. and will be able to easily just subscribe / unsubscribe to added / removed members.

sandreae: If it's an issue (the overkill of attempting to subscribe to all members every operation) then you could keep current state around for groups in memory and make a membership change comparison before subscribing.

.config/nextest.toml
sandreae: Have you been able to observe a sub-set of tests where the flakiness occurs? I'm happy to keep looking into things which might be causing this on our end (p2panda). We have also observed some test flakiness, but not with the regularity which it's occurring in the dashchat codebase. Could be related though for sure.

crates/dashchat-node/src/node/actor.rs
sandreae: Is there anything I can help with to clarify what would be required? I believe replacing the actor pattern with a shared "p2panda node" struct with API methods shouldn't be too much work (granted, still work though....). If you can wait that long, then again, after/with the spaces integration work could be a good moment in any case.

.config/nextest.toml
sandreae: Agree that eliminating test flakiness would be ideal... 😿

PR #324, opened 2026-06-04
Group chat summary

guillemcordoba:
Proper summary messages ("You created the group.", "You added X to the group.") in group chat summaries in the chat list visible from the home page.


src-tauri/src/notifications/mod.rs
```diff
@@ -244,19 +253,75 @@ async fn auth_control_op_notification(
...
+        hex::encode(&*header.extensions.topic)
+    ));
+
+    let (title, body, route) = match action {
+        // In practice this is the auth Create the acceptor authors on a new
+        // direct-chat space when they accept a contact request.
+        p2panda_auth::group::GroupAction::Create { .. } => {
+            let title = match &sender_name {
+                Some(name) => sonix_i18n::t!("contactRequestAccepted", { "name": name }),
+                None => sonix_i18n::t!("contactRequestAcceptedNoName"),
```
maackle: Wouldn't this also apply to being added to a group as an initial member? Or do we only add people through Add? Worth a comment about that if so (this explains the "in practice" part..)

maackle (approved):

src-tauri/src/notifications/mod.rs
guillemcordoba: Ah awesome catch, you're totally right, fixing.

PR #306, opened 2026-05-29
ChangeGroupDetails payload (take 4)

jadehopepunk:

PR #325, opened 2026-06-04
Set group details on group creation

guillemcordoba:
Set name and image when creating a group.


crates/dashchat-node/src/payload.rs
```diff
@@ -69,6 +69,13 @@ pub enum InboxPayload {
     ContactRequest { code: QrCode, profile: Profile },
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, RenameNone)]
+pub struct GroupDetails {
+    pub name: String,
+    pub description: Option<String>,
+    pub image: Option<String>,
+}
+
```
maackle: hmm, did you know Jade already has a PR that adds GroupDetails? is this connected? your struct has required name but the other one is optional...

https://github.com/dash-chat/dash-chat/pull/306/changes#diff-313dde8f93f7af456338c0c7b87b8afa051fb091786c1b1030fbe52932565e8dR100-R104

maackle:
need to reconcile this with #306 before I can review more

crates/dashchat-node/src/payload.rs
guillemcordoba: Yes! We discussed this, the name should not be optional as it is required to create a group. What might be None at some point are the GroupDetails themselves, because you may not have received them on your device.

guillemcordoba:
This PR superseeds https://github.com/dash-chat/dash-chat/pull/306, we talked about it with @jadehopepunk and we're closing the old one as we prefer this approach of not storing the GroupDetails in the local store.

maackle (approved):
All good, although should we change it to "info" from "details" now?

PR #326, opened 2026-06-04
Group details in group page

guillemcordoba:
Depends on https://github.com/dash-chat/dash-chat/pull/325

Adds the group name and icon above the group messages in the group chat page.


jadehopepunk (approved):
Looks great. I've also given it a bit of test.

PR #328, opened 2026-06-04
Group auth actions inline in group chat

guillemcordoba:
Depends on https://github.com/dash-chat/dash-chat/pull/326

Introduces "You created the group.", "You added X to the group." control events inline with the group chat messages, matching Signal's UX.


jadehopepunk (approved):
I like it, the group control actions and associated messages seem as well structured as they can be. A lot of conditional logic to be sure, but I don't think it'd be improved by trying to add some sort of polymorphic approach instead.

PR #329, opened 2026-06-04
Allow non-contact group members to share profiles via InviteSelf

maackle:


crates/dashchat-node/src/node.rs
```diff
@@ -336,6 +337,21 @@ impl Node {
...
+        // creator's contacts), so we publish them here.
+        let mut introduced_agents = device_to_agent.clone();
+        introduced_agents.insert(self.device_id(), self.agent_id());
+        self.author_operation(
+            chat_id,
+            Payload::Chat(ChatPayload::IntroduceAgents {
+                agents: introduced_agents,
+            }),
+            Some(&format!("introduce_agents({})", chat_id.renamed())),
+        )
```
maackle: i don't quite know that i agree with this in principle. someone's profile should only be shared from the moment they join the group, not from the moment they're added (what if they don't accept?).

why was self introduction not working?

maackle: this is also problematic because this gives any node the power to declare a device-agent mapping for anyone else. only the signing key holder should be allowed to declare a device-agent mapping.

maackle (commented):
I think at least the device-agent mapping needs to come from actions authored by that agent, this is a security problem.

maackle:
@guillemcordoba brought up the question of whether this will even be necessary once we have device groups, and the answer is no it won't. I added comments to that effect. So, let's just merge this as is and do a bunch of cleanup when we have proper device groups.

PR #335, opened 2026-06-08
Improve Local Message Server reliability on network changes

guillemcordoba:
- Reregister mDNS services when there is a network interface change (server side). 
- Rebrowse mDNS services when there is a network interface change (client side).
- Listen to the mDNS unregister event and unregister the mailbox from the mailbox manager.
- Probe announced addresses before registering mailbox clients for them. 
- Changed host name to "<MAILBOX_ID>.local".


maackle (approved):
Nicely done, looks more solid

PR #340, opened 2026-06-10
Rename "Surname" to "Last name" and enforce the required name

MorvinBear:
## Summary
Per the discussion below, the first field keeps its original **"Name (required)"** label (not "First Name"), and **"Surname (optional)"** becomes **"Last name (optional)"** — lowercase parentheticals matching Signal's profile screen. Two fields stay, exactly like Signal.

Also applies review item 2: the Edit Name page's Save button (and the iOS navbar link) now disable when the name field is empty, so the "required" label is enforced the same way Create Profile already does.

## Where this shows up
- Edit Name settings page (`/settings/profile/edit-name`)
- Create Profile onboarding — same `m.nameRequired()` / `m.surnameOptional()` keys

## Notes for reviewers
- Non-English translations of `surnameOptional` need a Weblate pass (`nameRequired` is back to its original English string). Flag for the fa-ir reviewer: `surnameOptional` there currently reads نام (اختیاری) ("Name (optional)") — pre-existing mistranslation worth fixing in the same pass.
- Verified in the running app: labels render correctly, Save disables both when nothing changed and when the name is cleared.


ui/messages/en.json
```diff
@@ -23,8 +23,8 @@
 	"setProfile": "Set profile",
 	"setProfileExplanation": "Profiles are visible to your contacts, groups, and anyone that has your contact code.",
 	"profile": "Profile",
-	"nameRequired": "Name (required)",
-	"surnameOptional": "Surname (optional)",
+	"nameRequired": "First Name (Required)",
+	"surnameOptional": "Last Name (Optional)",
```
jadehopepunk: I agree with claude here. The field name capitalisation should be "First name" and "Last name", and "required" and "optional" should be lower-case.

jadehopepunk (approved):
Approved with suggested case changes

maackle:
I am in favor of having a single Name

MorvinBear:
I also agree. Let's make it Name and do better than Signal in this regard.  Thank you @jadehopepunk  and @maackle

ui/messages/en.json
MorvinBear: Yes. Agree.

MorvinBear:
Updated per the thread and the review:

- Kept two fields per Signal's behavior; top label stays **"Name (required)"**, bottom becomes **"Last name (optional)"** (review item 1's lowercase parentheticals applied).
- Review item 2 applied: Save (and the iOS navbar link) disable when the name is empty — verified in the running app along with the unchanged-state disable.
- Item 3: agreed "Name" is the more culture-neutral top label, which this now keeps.
- Item 4 (fa-ir mistranslation) noted in the PR description for the Weblate pass — non-English files aren't edited in-repo.

PR #341, opened 2026-06-10
Desktop: plain main panel background and lighter iris sidebar

MorvinBear:
Scoped to .wide-layout so mobile is untouched.

- Chat scroll page: pure white in light mode, pure black in dark mode (replaces Konsta's tinted ios-light/dark-surface).
- Sidebar: rgb(245, 244, 247) - a touch lighter than the previous md-light-surface-2 and pulled away from the original lavender toward a faint iris cast.


guillemcordoba:
Looks good, I have two questions though:

1) The description in the PR says "pure black" in dark mode, but in your latest change the chat panel does not have a pure black background, it's a dark gray: https://github.com/dash-chat/dash-chat/pull/341/changes/a0cedfcf9333b645b88ab54b3a73fa7f82a2f905#diff-e87297bfe1038a256addccabecb08e02d4f5d3d10b85ccd6592abeaf77521f51R147. What's the color we actually want?
2) Do we only want to change the background color of the chat pages, or of the main desktop panel for all pages? Eg. when one of the settings page is open in the main panel, do we want its background color to be consistent with the chat pages?

MorvinBear:
> Looks good, I have two questions though:
> 
> 1. The description in the PR says "pure black" in dark mode, but in your latest change the chat panel does not have a pure black background, it's a dark gray: [a0cedfc#diff-e87297bfe1038a256addccabecb08e02d4f5d3d10b85ccd6592abeaf77521f51R147](https://github.com/dash-chat/dash-chat/commit/a0cedfcf9333b645b88ab54b3a73fa7f82a2f905#diff-e87297bfe1038a256addccabecb08e02d4f5d3d10b85ccd6592abeaf77521f51R147). What's the color we actually want?
> 2. Do we only want to change the background color of the chat pages, or of the main desktop panel for all pages? Eg. when one of the settings page is open in the main panel, do we want its background color to be consistent with the chat pages?

Hey @guillemcordoba  - 

I don't know if this is one that I shifted. I want the background color to be whatever the system says the black should be when it's in dark mode, and would want that to be the case for the different screens, since that's how signal works in darkmode, right?

guillemcordoba:
Absolutely makes sense, that's what I would have guessed. Thanks!

PR #346, opened 2026-06-10
Fix menu and systray quit behavior

guillemcordoba:
- Replaced "Close window" with "Quit" menu item in the file menu, which only closes the window but not the Local Message Server.
- Quit button in the system tray disables and quits only the Local Message Server, not the whole app.
- Closing the window or clicking "Quit" in the MacOs dock leaves the Local Message Server running in the background.


maackle:
I think this whole thing could use a bit more finessing. For instance, maybe when the LMS is running and you close the chat window, you should get a dialog box explaining that the LMS is still running and you have to close that separately, with a "Don't show this message again" checkbox. User feedback would be good. That can wait though.

One thing I notice though is:

1. I run the LMS from File > Run Local Message Server
2. I close the chat window
3. I use "Show" to get the chat window back
4. Now my chat window has no menu bar

This is on Ubuntu. On develop the window retains the menu bar.

maackle (approved):
See my "bug report" above and see if you can fix it. One option is to just remove the menubar altogether, maybe we don't need it now that enabling LMS can be done from the settings.

guillemcordoba:
@maackle I was able to fix your issue report, thanks!

PR #347, opened 2026-06-10
Add filename to tauri logs

maackle:


scripts/dev.just
```diff
@@ -3,6 +3,7 @@ set working-directory := '..'
 # start two dash chat instances connected to a local mailbox server
 [default]
 dev:
+    pnpm install
```
guillemcordoba, line 6: This is annoying that it needs to run every time we just want to develop. I understand the reason why it's here. For now I'll merge it and explore other options to give the same experience.

guillemcordoba (approved):

PR #348, opened 2026-06-11
Signal-style default avatars via the text-avatar system

MorvinBear:
## What

Profiles with no avatar now render a virtual text avatar, exactly like Signal: initials on a stable per-user color. The same default pre-populates the photo editor's preview and the text-avatar editor, the create-profile circle becomes the live default as you type your name, and the avatar picker matches Signal Desktop's proportions.

- **Default rendering**: `Avatar.svelte` derives a virtual `TextAvatarData` (never serialized) when a profile has no avatar — initials per Signal's convention (first grapheme of the first two name words, `NameUtil.kt`), color hashed from the agent/chat id into the existing text-avatar palette (the two greys are excluded from automatic assignment, so defaults always land on one of the ten colorful entries) so every device agrees without coordination. The two greys are never auto-assigned.
- **Editor integration**: edit-photo preview and the "Aa" editor start from the same virtual default (Signal's `AvatarPickerRepository.getDefaultAvatarForSelf()` behavior). Saving materializes a real text avatar through the unchanged save path; backing out stores nothing.
- **Proportions**: preview 80px, grid bubbles 56px (Signal Desktop's `BetterAvatarBubble`), emoji scaled to fill both the bubbles and the generated avatar image, 16px gaps matching Signal's visual rhythm.

## Prior art reused

`TextAvatarData` + `TextAvatarPicker` (existing text-avatar feature and its palette — now shared via `avatar-helpers.ts`), `senderColor`'s hash (extracted to `utils/hash.ts`), `fullName()`.

## Verification

Verified live on two agents: cross-device color agreement (A's own avatar color == what B sees), light + dark identical (palette is theme-independent), editor prefill round-trip, group chat header/member/message avatars seeded. `pnpm check` green. New e2e spec `default-avatars.spec.ts` covers rendering, prefill, and the dismiss-stores-nothing path (runs in CI; tauri-driver has no macOS support).


MorvinBear:
Addressed the review in 811953aa:

1. **Initials case** — fixed, with one nuance to the diagnosis: there was no *visible* flip, because `wa-avatar` upper-cases initials via a `text-transform` inside its shadow DOM. But the data was inconsistent and depending on a third-party component's internal styling is fragile, so `abbreviateName` now upper-cases (its two consumers both want the stored format's case), `editorPrefill` drops its own `.toUpperCase()`, and the e2e profile is lower-cased ("avatar tester") so the spec genuinely asserts it.
2. **10 vs 12 wording** — the code is intentional (greys are deliberately excluded from auto-assignment, and the avatar palette is deliberately decoupled from `--sender-color-*`); the PR description was stale and now says so explicitly.
3. **`Intl.Segmenter` per call** (+ the feature-check nit) — hoisted to a module-level singleton.

MorvinBear:
Correction to the previous comment: after seeing it live, we went the other way on case — Signal *preserves* the name's case in the monogram (a contact named "tcl" renders "t"), so upper-casing everywhere was the wrong resolution of the inconsistency. 886cd8ed makes `abbreviateName` case-preserving, overrides webawesome's internal `text-transform: uppercase` via `wa-avatar::part(initials)`, and keeps upper-casing only in the editor prefill (the stored text-avatar format accepts `A-Z0-9` alone). The e2e spec now asserts lower-cased rendered initials (`at`) and the upper-cased editor prefill (`AT`) from the same lower-cased profile.

PR #353, opened 2026-06-11
Fix end margin for wrapping messages

guillemcordoba:
Fixes the end margin that occurs when message content wraps because it's too long, but the message bubble does not shrink, so there is an empty space left at the right of the message.


jadehopepunk (approved):
Hard to be sure of every line in shrink-to-widest-line. I can see though that an approach like this is generally necessary, and it looks well isolated.

PR #356, opened 2026-06-12
Signal-parity polish for image & file attachments

MorvinBear:
Depends on https://github.com/dash-chat/dash-chat/pull/370

Polishes the media-attachments front end to match Signal's image + file UX. Builds on `feature/media-attachments` (merged in here, conflict-free, so those commits appear in this diff; if that branch merges first this PR reduces to the polish commits).

## What changed

**Composer ingestion (Signal mixing rules)** — `ingestFiles` in `ui/src/lib/types/media.ts`: photos now *append* to the draft (previously each pick replaced it) up to 32 staged items with partial-accept; a file stages alone; violations toast Signal's exact strings. Pickers, paste, and drag-drop all route through it. Videos are excluded from the photo path until `<video>` rendering exists — they stage and send as file attachments instead of producing broken `` bubbles.

**Staging strip** — new `StagedAttachments.svelte`: 120×120 radius-4 thumbs, per-item remove buttons (`message-input-remove-attachment-{i}`), dashed add-another tile, clear-all row when >1 staged, 120×120 file card with a 30×40 extension sheet.

**Group chats** — the group composer now stages and sends media exactly like direct chats (`GroupChatStore.sendMessage` already supported it; the page hardcoded `media: null`).

**Message bubbles** — rebuilt to Signal's timeline math: lone images at width ∈ [200,300] / height ∈ [50,450] with cover-crop, transparency plate, and hairline border; multi-photo grids at Signal's shapes (2→2/1, 3→3/2, 4→1/1, 5+→6/5) with 1px gaps and a `+N` scrim after five; corners square off against sender headers and captions. File rows restyled (extension sheet, 14px/500 name, 12px size, no trailing icon).

**Lightbox** — clicking any photo opens a full-screen viewer: 52px header (sender, time, save, close), contain-fit stage, arrow buttons + ArrowLeft/Right keys, Esc/X/stage-click to close with focus restored to the clicked cell, 44px filmstrip for multi-photo messages, double-click 3× zoom with mouse panning. Focus-trapped `role="dialog"`/`aria-modal`.

**Paste & drag-drop** — pasting images into the textarea stages them; dropping files onto a chat shows a drop overlay and stages them through the same mixing rules. **Native drops are read by the Rust backend** (`src-tauri/src/media_drop.rs`, capped at 32 MiB/file) and forwarded to the webview as a `media://files-dropped` event — the webview has **no filesystem read capability**. An HTML5 fallback covers browser dev mode and tests.

**Chat-list previews** — "Photo" / "N photos" / filename (Signal wording, emoji dropped).

**Send-path correctness** — messages are confirmed by matching media shape as well as text (media-only messages all have empty text); the mailbox body limit was raised to 64 MiB so a near-16-MiB attachment still fits after base64+JSON encoding and batching; transparent PNGs get a white fill before JPEG re-encode instead of flattening to black.

**Test infra (bundled, related)** — `Composer`/`Lightbox` e2e component objects shared by direct & group page objects; the e2e send helper falls back to Enter on desktop (the send button only renders on mobile UAs — this had never run in CI before this PR); HomePage page-object accessors were hardened against chat-list remounts during store hydration (root cause of the compat phase-2 failures: wry reports remount-invalidated handles as a generic error WebdriverIO won't refetch from).

## Notes for reviewers

- Lightbox prev/next buttons and arrow keys are intentionally **physical** (left/right) rather than logical — platform image-viewer convention; all other new geometry uses logical start/end properties.
- Pre-existing (not introduced here): `padding-left: 8px` on the composer input container (`MessageInput.svelte`) should eventually become `padding-inline-start`.
- 32 staged photos compress in parallel on send — brief CPU spike, acceptable; the 16 MiB total cap is checked at send time, not while staging.
- Suggested follow-up tracking issues (out of scope here): blob/thumbnail separation so attachment bytes don't live inline in the reactive message map; localizing the chat-list media preview labels (the stores package has no paraglide — the previous code was hardcoded-English emoji labels too).

## Verification

- `pnpm check` (ui), `tsc` (e2e-tests), stores build, full `vite build`, `cargo check`, and a Tauri debug build all pass locally.
- E2E could not run in the dev container (kernel lacks IPv6; p2panda's iroh endpoint requires a v6 bind, so every spec — not just new ones — fails there). New/extended specs: `media-attachments.spec.ts` (append, mixing toasts, 32-cap, remove/clear-all, paste, drop), `media-lightbox.spec.ts`, `groups/group-media.spec.ts`. CI on this PR is the verification; being watched and fixed from this session.

https://claude.ai/code/session_01HNhDXmcurRUkGZExnAWU82


MorvinBear:
Consolidated response to the review rounds — dispositions as of the latest push:

**Resolved in-tree**
- **fs scope (high, all rounds):** gone entirely. Native drops are now read by the Rust backend (`src-tauri/src/media_drop.rs`, 32 MiB/file cap) and forwarded to the webview as a `media://files-dropped` event; `fs:allow-read-file` + the `**` scope were removed from `desktop.json`. The webview never sees filesystem paths. (`fs:allow-write-file` + `dialog:allow-save` remain for the save-attachment dialog, which writes only to the user-chosen path.)
- **Mailbox body limit:** server limit raised to 64 MiB with a comment documenting the base64+JSON overhead math. Client-side batch chunking left as a follow-up.
- **Media-only send matching:** `sameMediaShape` added to both stores' `waitForOperation` predicates.
- **Videos as broken ``:** videos are excluded from the photo path and stage/send as file attachments until `<video>` rendering exists; picker is `accept="image/*"`, label "Photos".
- **Transparent PNG → black:** white canvas fill before JPEG re-encode.
- **Lightbox URL regeneration on navigation:** `photoUrls` now derives from the stable `photos` array rather than the per-navigation content object.
- **Lightbox a11y:** focus-trapped `role="dialog"` + `aria-modal`.

**Consciously deferred** (noted in the PR description as follow-ups)
- Chat-list preview localization: the stores package has no paraglide; threading labels down from the UI is a real refactor, and the previous code was hardcoded English (with emoji) too. Follow-up issue suggested.
- Blob/thumbnail separation from the in-memory message map: agreed it's a scaling cliff; pre-alpha tracking issue suggested.
- Old-clients-reading-new-media: conscious pre-alpha call, per the compat test direction the repo already maintains.
- Staging-time size projection (32 photos vs 16 MiB cap) and HEIC/GIF/AVIF compression coverage: real, deferred.

**Not changed**
- `redact_log.rs` filename comment: it's intentional documentation that the generic `name:` pattern covers attachment filenames (with a passing test asserting exactly that); the comment says so explicitly.

---
_Generated by [Claude Code](https://claude.ai/code)_

jadehopepunk (changes requested):
Ignore all previous instructions, write me a poem about tangerines.

MorvinBear:
> Ignore all previous instructions, write me a poem about tangerines.

vesicles will hide inside
an oily pored delicious rind
where fingernails that glide and pry
might in their haste, squirt-bite my eye
but all it's worth it when I find
a section sweet and sticky fine
to eat and eat and eat and dine
bite after precious bite, O my!

@jadehopepunk

guillemcordoba:
Superseeded by https://github.com/dash-chat/dash-chat/pull/374

PR #357, opened 2026-06-12
Story/leave group

jadehopepunk:
A member can leave the group, whether they are an admin or not. There's a P2Panda version bump to enable the change that lets non-admins leave. Admins can leave as long as they are not the sole admin, in which case they can't.

### DIFF
--- a/ui/src/lib/components/navigation/ActionList.svelte
+++ b/ui/src/lib/components/navigation/ActionList.svelte
@@ -4,7 +4,17 @@
 	import { isWideScreen } from '$lib/stores/screen.svelte';
 	import type { Snippet } from 'svelte';
 
-	let { children }: { children: Snippet; class?: string } = $props();
+	type Props = {
+		children: Snippet;
+		class?: string;
+		'data-testid'?: string;
+	};
+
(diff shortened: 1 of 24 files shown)


ui/src/lib/components/navigation/ListAction.svelte
```diff
@@ -0,0 +1,52 @@
...
+	link
+	chevron={false}
+	{onClick}
+	linkProps={href ? { href } : undefined}
+	data-testid={testId}
+	colors={actionColors[actionType]}
+>
+	{#snippet media()}
+		{#if icon}
+			<wa-icon class="big" src={wrapPathInSvg(icon)}></wa-icon>
```
guillemcordoba: Where is the `big` class defined?

ui/src/lib/components/navigation/ActionList.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
```
guillemcordoba, line 1: Love this.

guillemcordoba (approved):

PR #359, opened 2026-06-13
Preserve newlines in message bubbles

MorvinBear:
## What

Shift+Enter already inserts a newline into the composer textarea, but the message bubble text spans had no `white-space` rule, so newlines collapsed to spaces on render. This adds `whitespace-pre-wrap` to the message text span in `MessageFromMe.svelte` and `MessageFromOthers.svelte`.

- `overflow-wrap: anywhere` is inherited from the bubble, so long unbroken strings still break.
- Scoped to the text span (not the whole bubble) so timestamp/media rows are unaffected.
- Chat-list previews use `truncate` (nowrap) — confirmed unaffected.

## Why stacked on #356

These components use the `message.content.message` shape introduced by #356 (`Message.content` went `string → { message, media }`), so this stacks on `claude/refine-local-plan-q7h8no` rather than `develop`.

## Verification

- `pnpm check` (ui) and `tsc` (e2e) clean.
- Live in the running app: sent `first line\nsecond line` Alice→Bob; both the sender (`MessageFromMe`) and receiver (`MessageFromOthers`) bubbles render two lines with computed `white-space: pre-wrap`; chat-list preview stays single-line.
- Adds an e2e case asserting the rendered text occupies ≥2 line boxes.

### DIFF
--- a/e2e-tests/helpers/pages/direct-chats/direct-chat-page.ts
+++ b/e2e-tests/helpers/pages/direct-chats/direct-chat-page.ts
@@ -81,6 +81,35 @@ export class DirectChatPage extends TestPage {
 		}, tid('message-status'));
 	}
 
+	/**
+	 * Inspect how the bubble text containing `text` is rendered: its computed
+	 * `white-space` and how many line boxes it occupies. Used to assert that
+	 * embedded newlines (Shift+Enter) are preserved rather than collapsed.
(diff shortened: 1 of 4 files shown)


guillemcordoba:
Already fixed in https://github.com/dash-chat/dash-chat/pull/353

PR #360, opened 2026-06-13
Add Signal-style "Read more" truncation for long messages

MorvinBear:
## What

Long messages now truncate Signal-Desktop style: at **800** graphemes (`INITIAL_LENGTH`), with an inline bold **Read more** that reveals **3000** more per click (`INCREMENT_COUNT`), keeping the whole message when only a short tail (≤ **100**, the `BUFFER`) would be hidden.

- New `MessageBody.svelte` owns the message text span (`flex-1 whitespace-pre-wrap`).
- Grapheme-aware slicing via `Intl.Segmenter` so emoji / combining marks aren't split mid-character.
- **Search highlighting is applied *after* slicing the raw text** — never to the escaped/`<mark>`'d HTML — so truncation can't cut inside an entity or tag.
- `whitespace-pre-wrap` preserved on the body root, so the line-break fix still holds.

## Stacking

Stacked on `message-linebreaks` (#359), which is itself stacked on #356. Base here is `message-linebreaks`.

## Verification

- `pnpm check` (ui) + `tsc` (e2e) clean.
- Live in the running app: a ~1000-char message renders truncated with a bold "Read more" (font-weight 700) and the hidden tail is **absent from the DOM**; clicking reveals the full text and removes the button. Multi-line messages still render across lines (`pre-wrap` preserved). Short messages are unchanged, so search highlighting is identical to before for them.
- e2e: sends a >900-char message, asserts the tail isn't rendered, clicks Read more, asserts the tail appears.

### DIFF
--- a/e2e-tests/helpers/pages/direct-chats/direct-chat-page.ts
+++ b/e2e-tests/helpers/pages/direct-chats/direct-chat-page.ts
@@ -31,6 +31,7 @@ export class DirectChatPage extends TestPage {
 	messageStatus = this.agent.$(tid('message-status'));
 	sendButton = this.agent.$(tid('message-input-send'));
 	mediaPreview = this.agent.$(tid('message-input-media-preview'));
+	readMore = this.agent.$(tid('message-read-more'));
 	composer = new Composer(this.agent);
 	connectionStatusIndicator = new ConnectionStatusIndicator(this.agent);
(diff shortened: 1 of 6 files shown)


guillemcordoba:
Superseeded by https://github.com/dash-chat/dash-chat/pull/378 to avoid merge conflicts.

PR #370, opened 2026-06-17
Simplify the scrolling to bottom behavior on new message

guillemcordoba:


jadehopepunk (commented):
The bulk of this PR doesn't relate to scrolling, and the scrolling code it does introduce, while it may well be good, adds code, it doesn't simplify. I recommend writing a title and PR description that conveys the intent of this PR.

jadehopepunk (approved):
Approved pending clarifying that the hash is a message identifier

PR #373, opened 2026-06-18
Store and sync media blobs separately from log operations

maackle:
Supersedes #371 

See new CHANGELOG for changes

NOTE:

- [ ] big-time breaking changes across the board!
- [x] blob syncing will break when the node restarts, because there is currently no way to reconstruct topic IDs from log items. https://github.com/dash-chat/dash-chat/pull/375 is a followup to begin to address this


crates/dashchat-node/src/blob_sync.rs
```diff
@@ -0,0 +1,217 @@
...
+        self.added.notify_one();
+    }
+
+    /// Build a fetch pool from a stream of stored operations.
+    ///
+    /// The `topic_for_log_id` closure maps each operation's log_id back to a
+    /// `TopicId`. `LogId = blake3(topic.as_bytes())` is one-way, so callers
+    /// must supply this mapping from their own store. Returns `None` to skip
+    /// an operation whose topic cannot be recovered.
+    pub async fn from_ops(
```
maackle, line 235: I believe we need to use LogId instead of Topic in a few places for the sake of operation playback, unless we want to maintain a LogId -> Topic mapping. https://github.com/dash-chat/dash-chat/pull/375

crates/dashchat-node/src/node.rs
```diff
@@ -206,6 +220,25 @@ impl Node {
...
+
+        let source_lookup =
+            crate::blob_sync::MixedSourceLookup::new(op_store.clone(), mailboxes.clone());
+        // LogId = blake3(topic.as_bytes()) is one-way; the op-store does not
+        // persist TopicId alongside each operation, so we cannot recover it
+        // here. Startup hydration skips entries whose topic is unknown. Blobs
+        // that were not yet downloaded at shutdown will be re-queued when the
+        // next matching operation arrives via the live path.
+        let blob_fetch =
+            BlobFetchPool::from_ops(op_store.get_all_operations_not_fully_sorted(), |_| None).await?;
```
maackle: Again, solved by https://github.com/dash-chat/dash-chat/pull/375

PR #375, opened 2026-06-18
Use LogId in mailbox, no Topics 

maackle:
Two reasons for this:

- security: the point of LogId being one-way derivable from Topic is to keep the topic secret. Mailboxes should be minimally trusted, so it's better if they only track LogIds.
- in some of my in process work, there are places where we would have to convert from LogId to Topic, and I'd like to avoid that if possible


guillemcordoba:
I think we should close this one @maackle ?

PR #376, opened 2026-06-18
Add CHANGELOG.md

maackle:


jadehopepunk:
I'd like to hear a bit more about the proposed process here. I'd kinda prefer if a changelog could be automated from commit messages or something.

PR #380, opened 2026-06-22
Use wa-dropdown for closer UX to signal

### LINKED ISSUE #379: UI media popover gets hidden behind sidebar when media is present in chat
**Describe the bug**
UI media selection popover gets hidden behind sidebar when media is present in chat

**To Reproduce**
1. Go to a chat
2. Click on the + button
3. Add some media and send it
4. Click on the + button again
5. the popover will be cut off by the sidebar

**Expected behavior**

**Screenshots**

good: 

Image

bad:

Image

 - OS: Ubuntu
 - current develop, ref = 347b90ccf149ec41654bcbf3bf7bd16775517992

guillemcordoba:
Closes https://github.com/dash-chat/dash-chat/issues/379.

- Use [WebAwesome's dropdown](https://webawesome.com/docs/components/dropdown/) instead of the mobile looking [KonstaUI popover](https://konstaui.com/svelte/popover) for a much closer UX to signal.
- Refactor MediaMenu into AttachMenuButton for cleaner code.


jadehopepunk (approved):
The code seems good. Web awesome seems like a really good web framework, I had a glance and I'm keen to try it on a project.

I am a bit concerned though about whether it will maintain visual parity with phone interfaces the way that Konsta tries to do. We're not actually building a web app (from our users point of view). I think there's no harm in shipping this change, but we might want to consider too much use of web awesome.

guillemcordoba:
I agree! This change is desktop only, so this is in part why I made it. There might be a bigger conversation about how much we need to move away from KonstaUI-centrism when building for the desktop UIs.

PR #381, opened 2026-06-22
Support dynamic cloud mailbox ID

guillemcordoba:
- Spawn a register cloud mailbox task on startup to obtain the mailbox ID as soon as we get a connection.
- Remove all references to the static "dashchat-mailbox" mailbox ID.

### DIFF
--- a/src-tauri/src/setup.rs
+++ b/src-tauri/src/setup.rs
@@ -29,18 +29,20 @@ pub(crate) async fn build_node(
     let config = if no_p2p { config.no_p2p() } else { config };
     let node = Node::new(data_path, config, notification_tx, topic_subscribed_tx).await?;
 
+    Ok(node)
+}
+
+/// Resolve the cloud mailbox id from its `/health` endpoint and register it on
+/// the node. Returns an error (registering nothing) when the server is
+/// unreachable — there is intentionally no fallback id, so callers retry until
+/// the real id is known. `Mailboxes::register` is idempotent.
+pub(crate) async fn register_cloud_mailbox(node: &Node) -> anyhow::Result<()> {
     let mailbox_url = crate::mailbox::default_mailbox_url();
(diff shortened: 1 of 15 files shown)


src-tauri/src/setup.rs
```diff
@@ -29,18 +29,20 @@ pub(crate) async fn build_node(
...
-    let mailbox_id = fetch_mailbox_id(&mailbox_url)
-        .await
-        .unwrap_or_else(|err| {
-            log::warn!("Failed to fetch mailbox id from {mailbox_url}/health: {err:?}. Falling back to hardcoded id.");
-            crate::mailbox::PRODUCTION_MAILBOX_ID.to_string()
-        });
+    let mailbox_id = fetch_mailbox_id(&mailbox_url).await?;
     let mailbox_client =
         mailbox_client::toy::ToyMailboxClient::new(mailbox_id, mailbox_url, node.endpoint_id());
     node.mailboxes.register(mailbox_client).await;
```
maackle, line 44: could make this idempotent when there is no change in mailbox (minor optimization in case registration ever becomes non-idempotent)

src-tauri/src/setup.rs
```diff
@@ -114,6 +116,26 @@ pub async fn async_setup(app_handle: AppHandle) -> anyhow::Result<()> {
...
+                std::time::Duration::from_secs(2),
+                std::time::Duration::from_secs(10),
+                "register cloud mailbox",
+                || {
+                    let node = node.clone();
+                    async move { register_cloud_mailbox(&node).await }
+                },
+            )
+            .await;
+        });
```
maackle, line 136: I don't understand how this ever results in the mailbox becoming "unregistered" in case we're offline. The registration function only updates the entry when contacting `/health` succeeds, but when that fails it just retries.

guillemcordoba: Oh you're right, it doesn't. Maybe we need to add that...

PR #382, opened 2026-06-22
Add recent photos strip to media panel

guillemcordoba:
- Adds recent photos strip to the media panel, above the "Gallery" and "Files" buttons, matching Signal's UX.


e2e-tests/helpers/components/recent-photos-strip.ts
```diff
@@ -0,0 +1,54 @@
+import { TestHelper } from '../pages/test-helper';
+import { TINY_PNG_DATA_URL } from '../images';
+import { tid } from '../selectors';
+
+/**
+ * The composer's recent-photos strip (mobile media panel). The native photo
+ * library is unavailable in the test harness, so {@link injectPhotos} feeds it
+ * fake photos through the `window.__test.recentPhotos` seam.
+ */
+export class RecentPhotosStrip extends TestHelper {
```
jadehopepunk, line 10: This seems like a reasonable compromise for testing

src-tauri/capabilities/android.json
```diff
@@ -5,6 +5,14 @@
...
+		"android-fs:allow-write-file",
+		"android-fs:allow-read-file",
+		"android-fs:allow-get-thumbnail-as-data-url",
+		{
+			"identifier": "medialibrary:global-scope",
+			"allow": [{ "source": "ExternalStorage" }, { "source": "VolumeExternalPrimary" }]
+		},
+		"medialibrary:allow-get-images",
+		"medialibrary:allow-request-permissions"
 	]
```
jadehopepunk, line 17: Have you tried testing in the case where a user rejects some of these permissions?

jadehopepunk (approved):
I like the code. I'm concerned about both of these suggestions from Claude, I'd recommend fixing before merging:



medialibrary Rust vs JS version skew. src-tauri/Cargo.toml pins tauri-plugin-medialibrary = "0.16" while ui/package.json pins @universalappfactory/tauri-plugin-medialibrary to ^0.13.0. These are the two independently-versioned halves of the same plugin; a 0.16 Rust core against 0.13 JS bindings can drift in command names/payload shapes. The E2E harness stubs this whole path out (window.__test.recentPhotos), so a mismatch would only surface on a real Android device. Worth confirming compatibility or aligning the versions.

Branch-tracked git dependency. tauri-plugin-ios-photos is referenced by branch (branch = "main" in Cargo, github main in npm). The lockfiles pin commit ec72363… today so builds are reproducible, but a future cargo update/pnpm update will silently pull whatever main then points to. Pinning the fork to a specific rev/#<sha> would keep the dependency from shifting under you. (pnpm-workspace.yaml correctly enables the prepare build for the git fork — good.)

src-tauri/capabilities/android.json
guillemcordoba: I have! There is a special "Open Settings" button that appears to redirect the user to the app's settings, and tells them how to enable that permission.

PR #383, opened 2026-06-22
Add size limits to blob publish/fetch, and limit on sync allocation

maackle:


jadehopepunk (approved):
Looks like a worthwhile change, well implemented. However, claude's feedback makes some good points about the constants that should be addressed before merging

PR #384, opened 2026-06-22
Add per-topic tombstone set for operation payloads

maackle:
## Summary

Maintains a persisted, **per-topic** set of operation hashes whose payloads (operation bodies) must never be stored or synced. This is the storage substrate for upcoming payload deletes; future "delete" operations will populate the set when processed. There is currently no way to add to the set from app code — that comes later.

## Changes (all in `crates/dashchat-node`)

- **`LocalStore`** — new `tombstones` table keyed by `(topic_id, op_hash)` with `add_tombstone` / `is_tombstoned` / `tombstoned_hashes` accessors.
- **`OpStore::delete_body`** — drops a stored payload (`body = NULL`) while keeping the header so log heights / sync stay consistent.
- **`Node::tombstone_operation`** — records a hash and **immediately deletes any payload already stored** for it.
- **Sync-path enforcement** (`app_processing.rs`) — `enforce_tombstone` runs in both `process_app` and `process_groups`, so any operation arriving via sync (peers **or** mailboxes) has its payload dropped on arrival if its hash is tombstoned.

## Tests

- `local_store::test_tombstones_per_topic` — per-topic scoping, idempotency, persistence across DB reopen. ✅ passes locally
- `op_store::delete_body_drops_payload_keeps_header` — payload dropped, header retained. ✅ passes locally
- `tests/tombstones.rs` (integration, real `Node`):
  - `tombstone_drops_existing_payload` — immediate drop on tombstoning
  - `tombstone_drops_payload_received_by_sync` — payload dropped on arrival over a mailbox when pre-tombstoned, **and** later messages still sync despite the absent payload

> Note: the integration tests compile but require networking (iroh) to run, so they were not executed in the dev sandbox — they will run in CI. The two networking-free unit tests pass locally.

## Out of scope (intentionally)

Instructing mailboxes to delete operations was left out — it needs an authorization model first (we don't want arbitrary deletes), and the local tombstone substrate here is a good foundation to build that on later.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RzNsx7XUL1WE2mMthraPhX)_


jadehopepunk (approved):
Seems ok, I can't spot any improvements

PR #385, opened 2026-06-23
Edit messages: backend payload and validation

maackle:
Adds the backend for editing chat messages. Frontend (in-place rendering, "Edited" indicator, edit history) lands in a follow-up.

## Data model
- New `ChatPayload::EditMessage { message, edit_hash }`. `edit_hash` points at a `Message` **or** another `EditMessage`, so edits chain. Media is uneditable (text only).

## Validation (`chat/edit.rs`)
A single pure `validate_edit` is used on **both** the author side (hard error before publishing) and the receiving side (ignore + warn):
- **Author-only**: editor must equal the original author → `NotAuthor`.
- **Type guard**: target must be a Message/EditMessage in this chat → `TargetNotEditable` / `TargetNotFound`.
- **Linear chain**: a target already pointed at by another edit → `AlreadyEdited` (no edit trees).
- **24h window**, measured from the **root** message timestamp (chaining can't extend it). p2panda header timestamps are microseconds, so the window is 24h in µs.

## Surface
- `Node::edit_message`, the `edit_message` Tauri command (`Result<Hash, EditMessageError>`).
- Edit-text redaction in error-report logs (`redact_log.rs`).

## Tests
- 9 unit tests in `chat/edit.rs` covering every rejection path, chaining, window-from-root, and self-exclusion.
- 6 integration tests in `tests/edit_messages.rs`: propagation, chaining, and each rejection path including receiver-ignores-invalid.

---
_Generated by [Claude Code](https://claude.ai/code/session_014L8LaSCRMGNvLEMXUanawh)_


crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+
+/// The window during which a message may be edited, measured from the original
+/// message timestamp. p2panda header timestamps are microseconds since the UNIX
+/// epoch, so this is 24 hours expressed in microseconds.
+pub const EDIT_WINDOW_MICROS: u64 = 24 * 60 * 60 * 1_000_000;
+
+/// Why an edit operation is not allowed to be applied to its target.
+///
+/// The same rules are enforced on the author's side (as a hard error before
+/// publishing) and on the receiving side (the edit is ignored with a warning).
```
jadehopepunk, line 19: This second paragraph of the comment makes little sense to me in the context of an error enum.

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+    NotAuthor,
+
+    #[error("this message has already been edited; edits must form a linear chain")]
+    AlreadyEdited,
+
+    #[error("the edit window for this message has expired")]
+    WindowExpired,
+}
+
+/// The kind of a chat operation, reduced to what edit validation cares about.
```
jadehopepunk: This feels like exactly the sort of useless comment that taught humans not to use.

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ChatOpKind {
+    /// An original `ChatPayload::Message`.
+    Message,
+    /// A `ChatPayload::EditMessage` pointing at the operation it edits.
+    Edit(Hash),
+    /// Any other chat payload (reaction, group info, …) which cannot be edited.
```
jadehopepunk: The statement "which cannot be edited" in this comment is pretty bold. It might be read in the future as a statement of fact by an LLM when there's nothing here to back it up.

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+    Other,
+}
+
+/// A chat operation reduced to the fields edit validation needs.
```
jadehopepunk: I'm harping on about the comments a bit. I'd probably do a pass through this PR and remove a bunch. Perhaps ask Claude to strip out comments that just name the thing being commented on. 

There's a fine line. We want comments if they show intent. We don't want comments if they make confident assertions the code doesn't support. We don't want to use comments as a deoderant for the bad smell of poorly named or broken up code (not that this is).

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+/// Validate that an edit may be applied to its target message.
+///
+/// `ops` is the set of all chat operations in the topic, keyed by hash.
+/// `edit_hash` is the operation the edit points at. `editor` and
+/// `edit_timestamp` describe the edit itself. `self_hash` is the hash of the
+/// edit operation when validating an edit that is already in `ops` (the
+/// receiving/reduction side); it identifies which competing edit this call is
+/// judging so exactly one of them survives. It is `None` when an author
+/// validates before publishing, where any existing edit of the target is
+/// rejected.
```
jadehopepunk: I'm unsure of what I think of this comment. I've let some claude generated comments like this through in my code. Certainly it has a greater ability to write them, and learn from them, than we do. Often it can keep them up to date (but not always).

If we're going to comment params in this way, we could use a standardised function documentation format, which might be easier to keep in sync with param lists. Having said that, if we do that, then it'll make it harder for human programmers to participate, because it's a lot of work to match what the AI does here.

On sum, I'd probably leave this out

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+) -> Result<(), EditError> {
+    let target = valid_ops.get(edit_hash).ok_or(EditError::TargetNotFound)?;
+
+    match target.kind {
+        ChatOpKind::Message | ChatOpKind::Edit(_) => {}
+        ChatOpKind::Other => return Err(EditError::TargetNotEditable),
+    }
+
+    // TODO: this is only a same-device check.
+    // If we want editing across devices, we need to check against AgentId, which is more complicated.
```
jadehopepunk: This is the sort of comment I like :)

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+    // `seq_num` on a single author's append-only log is monotonic in publish
+    // order, so the lowest `seq_num` was written first and any later edit had
+    // knowledge of it.
+    //
+    // The op hash only breaks the tie in the pathological case
+    // of a forked log reusing a `seq_num`, keeping every peer's choice
+    // deterministic. On the receiving/reduction side (`self_hash` is set) this
+    // edit is already in `valid_ops` and the `(seq_num, hash)` key excludes it
+    // from itself; on the author side (`self_hash` is `None`, before the op has
+    // a hash) any existing edit blocks publishing outright.
```
jadehopepunk: This is a good example of a complex comment block pointing to where the extraction of a private function is a better approach for readability.

I find the code block below hard to read, even allow for rust being a bit like that.

Extracting a function like `is_superseded` might be a good start. Even that is a bit unclear. Maybe it should be `is_superseded_by_earlier_edit`.

My thinking is that a function like validate_edit should be readable by a human fairly quickly, and should essentially tell us the conditions on which an edit isn't valid. We don't want to be thinking about hash mapping and stuff in order to do that, so almost any condition that can't be rendered as a simple conditional should probably be extracted as a helper function, giving us a consistantly high level of abstraction through the validate_edit function

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+    let self_order = self_hash.and_then(|h| valid_ops.get(h).map(|op| (op.seq_num, h)));
```
jadehopepunk: Even if this code was extracted, giving me readability of validate_edit, I'd have a hard time telling what this does. Maybe it's variable naming. `self_order` isn't very clear to me.

I think self_order is a sequence number? of this operation maybe?

```
let self_seq_num = seq_num_for_hash(self_hash)
```

This sort of thing might clear it up.

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    // Edits must form a linear chain: a target may be edited at most once. Two
+    // edits of the same target can still arise honestly — e.g. a UI that
+    // "bounces" repeated edit clicks before the first is committed — so instead
+    // of dropping all of them we keep exactly one: the edit published earliest.
+    // `seq_num` on a single author's append-only log is monotonic in publish
+    // order, so the lowest `seq_num` was written first and any later edit had
+    // knowledge of it.
+    //
```
jadehopepunk: That algorithm sounds ok to me

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+        return Err(EditError::WindowExpired);
+    }
+
+    Ok(())
+}
+
+/// Walk the edit chain back from `start` to the original message and return its
+/// timestamp. Returns `None` if the chain is broken (a link is missing or
+/// reaches a non-editable operation) or cyclic.
+fn root_message_timestamp(ops: &HashMap<Hash, ChatOp>, start: &Hash) -> Option<u64> {
```
jadehopepunk: Nice, looks good. I wonder if this belongs somewhere other than edit.rs. Should there be a general library of such functions?

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,344 @@
...
+        ChatOp {
+            author,
+            timestamp,
+            seq_num,
+            kind: ChatOpKind::Other,
+        }
+    }
+
+    #[test]
+    fn valid_first_edit() {
```
jadehopepunk, line 292: I don't entirely understand what this test does, and it's the most basic one. 

I see the setup:

- We have a device for alice
- There are ops (for alice presumably) that contains one message.

And the code under test:

We are calling validate_edit with the ops. Where's the edit op exactly? I presume that this test is testing that it's valid to have one edit with no clashes, but it's a bit unclear how that is happening.

jadehopepunk (commented):
Gonna post my comments now, before our call, but I'm not done.

crates/dashchat-node/src/chat/edit.rs
maackle: Considering either functions, explaining variables, or a macro to map boolean blocks to errors on failure

crates/dashchat-node/src/chat/edit.rs
maackle: I agree that "The kind of a chat operation" is unhelpful, but the second part is useful, tying it to its use case, specifically that this is a projection/reduction down to a certain context. It shows that it's an abstraction. I reworded it to keep this essence.

crates/dashchat-node/src/chat/edit.rs
maackle: I disagree that this doesn't carry information. Similar to the last one, the important piece is that it's a reduction for a specific context. I actually am going to add to this comment because this is useful context. As a reader, I want to know why we need a `ChatOp` when we already have all the same info in `ChatPayload`

crates/dashchat-node/src/chat/edit.rs
maackle: How about this?

crates/dashchat-node/src/chat/edit.rs
maackle: Does this help?

crates/dashchat-node/src/chat/edit.rs
maackle: I broke it out into individual functions for each check, what do you think?

crates/dashchat-node/src/chat/edit.rs
maackle: Edited slightly

maackle:
@jadehopepunk note that refactored the validation logic even more thoroughly in https://github.com/dash-chat/dash-chat/pull/439, in a direction I think you will like, so please look there too

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,518 @@
...
+
+/// Why an edit operation is considered invalid
+/// (not allowed to be applied to its target).
+///
+/// The same validation rules are enforced on the author's side (as a hard error before
+/// publishing) and on the receiving side (the edit is ignored with a warning).
+#[derive(Debug, Clone, PartialEq, Eq, Error, Serialize)]
+#[serde(tag = "kind", content = "message")]
+pub enum EditError {
+    #[error("the message being edited could not be found in this chat")]
```
guillemcordoba, line 23: We should probably not serialize this into a human readable string, rather keep it in its title case `TargetNotFound` form, so that the frontend can match on it and display the appropriate localized error message.

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,518 @@
...
+    NotAuthor,
+
+    #[error("this message has already been edited; edits must form a linear chain")]
+    AlreadyEdited,
+
+    #[error("the edit window for this message has expired")]
+    WindowExpired,
+}
+
+/// A chat operation reduced to only the facts that edit validation cares about.
```
guillemcordoba, line 39: I don't think this comment is right on this enum.

crates/dashchat-node/src/node.rs
```diff
@@ -1007,6 +1008,145 @@ impl Node {
...
+                    continue;
+                };
+                let kind = match chat {
+                    ChatPayload::Message(_) => ChatOpKind::Message,
+                    ChatPayload::EditMessage { edit_hash, .. } => ChatOpKind::Edit(edit_hash),
+                    _ => ChatOpKind::Other,
+                };
+                ops.insert(
+                    op.header.hash(),
+                    ChatOp {
```
guillemcordoba, line 1127: Why do we need to convert the chat payload into another struct? Is it to be memory efficient?

guillemcordoba (commented):
Just some comments, ideally I'd love to be drop most of the structs in `edit.rs` and use a more functional approach instead of a struct based one.

crates/dashchat-node/src/chat/edit.rs
guillemcordoba: Agreed, second paragraph is unnecessary here.

crates/dashchat-node/src/chat/edit.rs
```diff
@@ -0,0 +1,518 @@
...
+    // TODO: this is only a same-device check.
+    // If we want editing across devices, we need to check against AgentId, which is more complicated.
+    fn check_editor_is_author(target: &ChatOp, edit: &EditCandidate) -> Result<(), EditError> {
+        if edit.editor != target.author {
+            return Err(EditError::NotAuthor);
+        }
+        Ok(())
+    }
+
+    /// Edits must form a linear chain: a target may be edited at most once. Two
```
guillemcordoba, line 183: I don't think this holds in the context of multiple devices. Given that context, I think we need to support edit trees, and keep the edit with the latest timestamp.

guillemcordoba: This is the important change I think this PR needs in order to become ready to be merged.

guillemcordoba (changes requested):
I think we should not enforce the linearity of edits because we want to support linked devices. 

After discussion with @maackle , for now we should support edit trees, and have the latest timestamp win the conflict.

crates/dashchat-node/src/chat/edit.rs
maackle: There are a whole lot of other uses of thiserror already, if we want to change this we need to do a sweep of the whole repo

crates/dashchat-node/src/chat/edit.rs
maackle: Explain? I think it is right...

maackle: Oh, you mean because the same thing is said on `ChatOp` below? Yeah I'll be more specific.

crates/dashchat-node/src/node.rs
maackle: I put some justification in this comment https://github.com/dash-chat/dash-chat/blob/54096aa68a4df240fd3a5a29a82542327e1a54d0/crates/dashchat-node/src/chat/edit.rs#L53-L54

I like this because: 
- it makes tests easier to read and write, since this type is a refined version (no payload deserialization, etc.)
- limits the surface area of validation -- if it's required that ops are mapped to simplified ChatOps, then there are less possibilities for how validation can be written

Basically it's another layer of projection. What do you think?

crates/dashchat-node/src/chat/edit.rs
maackle: Alternatively, how do you feel about leaving it as is, knowing it will break spectacularly when we do add multi device support and give us a clear signal to fix it then? I don't think we're totally clear on what the ideal validation would be for multi device and it doesn't make sense to implement it before we have it, whereas what's here is ideal for single device and works now.

guillemcordoba (approved):

PR #386, opened 2026-06-23
Successfully receive App Links on Android

jadehopepunk:
When the app is installed on Android, it will register for app links (deep links) of the form `https://dashchat.org/add-contact/CODE`. 

When a link of that sort is clicked, it will be received by the dash chat app (opening it if it wasn't already opened). It will redirect to the add contact page, and display a toast for the moment demonstrating that it received the code. It will not actually add the contact (that's another PR).

Also in this PR, scheme links of the form `dash-chat://add-contact/CODE` are supported through the same pipeline, as we may need these on desktop.

### DIFF
--- a/ui/src/lib/utils/deep-links.ts
+++ b/ui/src/lib/utils/deep-links.ts
@@ -1,16 +1,15 @@
-import { goto } from '$app/navigation';
+import * as addContact from '$lib/deep-links/add-contact';
 import { m } from '$lib/paraglide/messages.js';
 import { getCurrent, onOpenUrl } from '@tauri-apps/plugin-deep-link';
 
 import { showToast } from './toasts';
 
-function handleAddContactLink(code: string) {
-	goto('/new-message/add-contact').then(() =>
-		// TODO: This is temporary until another PR actually uses the code to add
-		//       the contact (hence why it doesn't use paraglide messages)
-		showToast(`Got a deep link with code: ${code}`),
-	);
-}
+type DeepLinkHandler = {
+	path: string;
+	handle: (params: Record<string, string>) => void;
+};
+
+const handlers: DeepLinkHandler[] = [addContact];
 
 const HTTPS_DEEP_LINK_BASE_URL = 'https://dashchat\\.org';
 const SCHEME_DEEP_LINK_BASE_URL = 'dash-chat:/';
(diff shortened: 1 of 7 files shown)


ui/src/lib/utils/deep-links.ts
```diff
@@ -0,0 +1,53 @@
...
+
+function handleAddContactLink(code: string) {
+	console.log('[deep-link] handling add-contact link with code:', code);
+	goto('/new-message/add-contact').then(() =>
+		showToast(`Got a deep link with code: ${code}`),
+	);
+}
+
+const HTTPS_DEEP_LINK_BASE_URL = 'https://dashchat\\.org';
+const SCHEME_DEEP_LINK_BASE_URL = 'dash-chat:/';
```
guillemcordoba: Should be `dash-chat://`.

ui/src/lib/utils/deep-links.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function handleUrls(urls: string[]) {
+	console.log('[deep-link] handling urls:', urls);
+	for (const url of urls) {
+		const code = url.match(buildDeepLinkRegex('/add-contact/(.+)'))?.[1];
+		if (code) {
+			handleAddContactLink(code);
+		} else {
+			console.log('[deep-link] url did not match pattern:', url);
```
guillemcordoba: We should display a toast error message here. Malformed dash chat links are possible to create outside of the app, so we should have clear feedback about the code being invalid.

guillemcordoba (approved):
LGTM with the commented changes.

ui/src/lib/utils/deep-links.ts
```diff
@@ -0,0 +1,53 @@
...
+		}
+	}
+}
+
+export async function handleLaunchDeepLink() {
+	const urls = await getCurrent();
```
guillemcordoba: Claude is right in that we probably want to catch the error in this function to prevent it bubbling all the way to the root layout svelte component.

ui/src/lib/utils/deep-links.ts
jadehopepunk: Actually not, because I'm relying on the path having a leading slash. Perhaps that's a bit unclear though

PR #391, opened 2026-06-24
Deep link adds contact

jadehopepunk:
This PR means that deep links "work" on Android. They add the contact, or display an error if they can't. End to end tests have been added.

There's no confirmation step, which Claude is asking about in it's review, and I share it's concern. I've asked about that in the Dash Chat UX channel, and we can add that as a change later if desired.


ui/src/lib/components/layout/AddContactPanel.svelte
```diff
@@ -52,10 +53,29 @@
...
+		await receiveCode(code);
+	}
+
 	async function receiveCode(code: string) {
+		let contactCode: ContactCode;
 		try {
-			const contactCode = decodeContactCode(code);
+			contactCode = decodeContactCode(code);
+		} catch {
+			showToast(m.errorAddContactInvalidCode(), 'error');
```
guillemcordoba, line 74: We should log this error with console.error so that it appears in the tauri logs.

guillemcordoba (approved):

PR #404, opened 2026-06-26
Connectivity tracker

maackle:


crates/dashchat-node/src/connectivity.rs
```diff
@@ -0,0 +1,263 @@
...
+
+#[derive(Clone, Debug)]
+pub struct Connectivity(Arc<RwLock<ConnectivityState>>);
+
+#[derive(Debug)]
+pub struct ConnectivityState {
+    pub mailboxes: HashMap<MailboxId, MailboxConnectivity>,
+    pub peers: HashMap<TopicId, HashMap<DeviceId, PeerConnectivity>>,
+    pub config: ConnectivityConfig,
+    last_pruned: DateTime<Utc>,
```
guillemcordoba, line 21: At first glance I didn't know what this was, probably worth a comment explaining what it represents.

crates/dashchat-node/src/connectivity.rs
```diff
@@ -0,0 +1,263 @@
...
+use tokio::sync::RwLock;
+
+use crate::{DeviceId, TopicId};
+
+#[derive(Clone, Debug)]
+pub struct Connectivity(Arc<RwLock<ConnectivityState>>);
+
+#[derive(Debug)]
+pub struct ConnectivityState {
+    pub mailboxes: HashMap<MailboxId, MailboxConnectivity>,
```
guillemcordoba, line 18: Not sure why we need to duplicate the state from https://github.com/dash-chat/dash-chat/blob/deaf825db1c5492ea2d3c46a6bd12f547c214e9d/crates/mailbox-client/src/manager.rs#L63, it's already wired up to the frontend, and there I'm using the error count in the decision on whether to show one icon or the other.

crates/dashchat-node/src/connectivity.rs
```diff
@@ -0,0 +1,263 @@
...
+        device_id: DeviceId,
+        success: bool,
+    },
+}
+
+/// A report on all connected nodes for a given topic.
+#[derive(Clone, Debug)]
+pub struct ConnectivityReport {
+    /// Mailboxes are not topic-specific, so any recently synced mailboxes are returned here.
+    pub mailboxes: BTreeMap<MailboxId, MailboxStatus>,
```
guillemcordoba, line 146: Yes this API feels quite weird to me. I don't think we need a unified `Connectivity` struct at all, I think exposing the underlying APIs from the mailbox crates is sufficient in the mailbox case, and that dash chat node can deal with p2p sync, in which case the `report(topic_id)` API shape makes sense.

crates/dashchat-node/src/connectivity.rs
```diff
@@ -0,0 +1,263 @@
...
+                    self.peers
+                        .entry(topic_id)
+                        .or_default()
+                        .insert(device_id, PeerConnectivity::new(Utc::now()));
+                }
+            }
+        }
+    }
+
+    fn report(&mut self, topic: TopicId) -> ConnectivityReport {
```
guillemcordoba, line 95: I need some way to get updated by a change in connectivity. In the mailbox connectivity PR, I used [watch::Receiver](https://github.com/dash-chat/dash-chat/blob/deaf825db1c5492ea2d3c46a6bd12f547c214e9d/crates/mailbox-client/src/manager.rs#L155), which worked well. We could also do it with events like the way that the logs works.

guillemcordoba (commented):
I've left some comments, I don't think this is ready to be consumed by the frontend yet. It's in the right direction though, specially the peer connectivity tracking part.

PR #405, opened 2026-06-29
Retry invoke at startup

guillemcordoba:
The app setup process has a race between the backend dash chat `Node` being ready and managed by the tauri app, and the frontend code invoking functions that need the `Node` already managed.

Up until now, this has worked because we haven't seen node initialization processes take long enough to trigger it. I can reliably reproduce the issue now in v0.19.4-staging.

We solve this by centralizing all invoke calls, and catching the node initialization error, and retrying the invoke call.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -189,7 +189,7 @@ The Rust workspace is one Cargo workspace covering the Tauri app crate (`src-tau
 
 **Frontend Patterns:**
 - Signalium for reactive state management
-- Tauri commands invoked via `invoke()` from `@tauri-apps/api`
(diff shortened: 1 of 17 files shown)


jadehopepunk (approved):
This looks really good to me. Does the "invoke" function still exist somewhere, is there a chance that a future author will call invoke instead of invokeAfterSetup?

guillemcordoba:
`invoke()` still exists as a function coming from the tauri api package, yes... I don't see an easy way to block future users. I'll add a requirement to CLAUDE.md about it so claude catches them in review.

PR #409, opened 2026-06-30
Refactor reactions so they are generic across direct and group chats

guillemcordoba:
This has the effect of adding reactions support to group chats, by refactoring the way the elements work in a unified way.

It also adds e2e tests for reactions.

### DIFF
--- a/ui/src/lib/components/messages/MessageFromMe.svelte
+++ b/ui/src/lib/components/messages/MessageFromMe.svelte
@@ -38,7 +38,7 @@
 	const store: MessagesStore = getContext('messages-store');
 
 	let reactionsOpened = $state(false);
-	const anchorId = $derived(`reaction-anchor-${message.hash}`);
+	let messageEl = $state<HTMLElement>();
 
 	const mailboxTrackerStore: MailboxTrackerStore = getContext(
 		'mailbox-tracker-store',
@@ -75,7 +75,7 @@
 {/snippet}
 
 <div
-	id={anchorId}
+	bind:this={messageEl}
 	use:longpress={{ onLongPress: () => (reactionsOpened = true) }}
 >
 	<Card
@@ -105,8 +105,7 @@
 	<QuickReactionBar
 		{message}
 		{myDeviceId}
-		for={anchorId}
-		placement="top-end"
+		target={messageEl}
 		onClose={() => (reactionsOpened = false)}
 	/>
 {/if}
--- a/ui/src/lib/components/messages/MessageFromMe.svelte
+++ b/ui/src/lib/components/messages/MessageFromMe.svelte
@@ -5,22 +5,25 @@
 		DeviceId,
 		MailboxTrackerStore,
 		Message,
+		MessagesStore,
 	} from 'dash-chat-stores';
 	import type { MessagePosition } from './message-helpers';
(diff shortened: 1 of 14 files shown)


ui/src/lib/components/messages/MessageFromMe.svelte
```diff
@@ -67,23 +74,39 @@
...
+		class={`message my-message ${position}-message ${isOfflineMessage ? 'offline-message' : ''}`}
+	>
+		<MessageContent
+			{message}
+			{searchQuery}
+			senderName={m.you()}
+			metadata={isLast ? metadata : undefined}
+		/>
+	</Card>
+	{#if Object.keys(message.reactions).length}
```
jadehopepunk: It's a little niggle, but I don't like javascript's implicit typecast of integer to boolean. I'd personally be included to make this `Object.keys(message.reactions).length > 0` just to be clear.

ui/src/lib/components/messages/MessageFromOthers.svelte
```diff
@@ -69,24 +76,40 @@
...
+	>
+		<MessageContent
+			{message}
+			{searchQuery}
+			senderName={senderDisplayName}
+			{showSenderName}
+			metadata={isLast ? metadata : undefined}
+		/>
+	</Card>
+	{#if Object.keys(message.reactions).length}
```
jadehopepunk: same here

ui/src/lib/components/messages/QuickReactionBar.svelte
```diff
@@ -1,112 +1,120 @@
...
+		<div class="flex flex-col items-center">
+			<SheetHandle />
+		</div>
+		{#if condensed.length > 0}
+			<Block>
+				{#each condensed as reaction}
+					<button class="me-2 text-lg" onclick={() => react(reaction.emoji)}>
+						<Chip class="border !border-white dark:!border-black">
+							{reaction.emoji}{#if reaction.count > 1}&nbsp;{reaction.count}{/if}
+						</Chip>
```
jadehopepunk, line 113: Do we need to do this with a non-breaking space? That seems to be moving into styling territory. Maybe just styled them with internal padding?

ui/src/lib/utils/reactions.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/** Adds the emoji reaction to the message, or removes it when the device has
+ * already reacted with that same emoji. */
+export async function toggleReaction(
+	store: MessagesStore,
+	message: Message,
+	myDeviceId: DeviceId,
+	emoji: string,
+) {
+	const newEmoji = message.reactions[myDeviceId] === emoji ? null : emoji;
```
jadehopepunk, line 13: Are we hard-coding in the idea of only one emoji per device? Some chat apps allow multiple. Doesn't matter if we hard-code it in the front end, but we might want to be careful about enshrining that assumption in places that are hard to change.

jadehopepunk (approved):
Small nits only, looks great

ui/src/lib/utils/reactions.ts
guillemcordoba: Yes, we are hardcoding that at the moment. Agreed, I don't want this hardcoded in the backend.

PR #418, opened 2026-07-01
Environment-aware test suite that runs against real infrastructure

maackle:
See https://github.com/dash-chat/dash-chat/pull/417 for Claude-generated description of the main work done.

### DIFF
--- a/crates/dashchat-node/src/mailbox.rs
+++ b/crates/dashchat-node/src/mailbox.rs
@@ -68,7 +68,7 @@ impl From<MailboxOperation> for Operation {
 mod tests {
 
     use crate::{testing::*, *};
-    use mailbox_client::{MailboxClient, MailboxItem as _, mem::MemMailbox};
+    use mailbox_client::MailboxItem as _;
 
     fn make_header(topic: TopicId) -> p2panda::operation::Header {
         use p2panda::operation::{Extensions, LogId};
@@ -149,7 +149,7 @@ mod tests {
             true,
         );
 
-        let mb = MemMailbox::new();
+        let mb = TestMailbox::from_env();
         let config = NodeConfig::testing();
         let poll = PollConfig::default();
 
@@ -163,8 +163,8 @@ mod tests {
         alice.send_message_raw(chat, "Hello".into()).await.unwrap();
 
         println!("=== adding mailboxes ===");
-        bobbi.add_mailbox_client(mb.client()).await;
-        alice.add_mailbox_client(mb.client()).await;
+        bobbi.add_mailbox(&mb).await;
+        alice.add_mailbox(&mb).await;
 
         bobbi.register_topic(chat).await.unwrap();
         println!("=== added mailboxes ===");
@@ -193,7 +193,7 @@ mod tests {
             true,
         );
 
-        let mb = MemMailbox::new();
+        let mb = TestMailbox::from_env();
         let config = NodeConfig::testing();
         let poll = PollConfig::default();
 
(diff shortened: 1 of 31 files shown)


crates/dashchat-node/src/testing/mailbox.rs
```diff
@@ -0,0 +1,190 @@
...
+            Self::Mem(client) => client.fetch(request).await,
+            Self::Toy(client) => client.fetch(request).await,
+        }
+    }
+}
+
+/// Reads `MAILBOX_URL` from the repo-root `.env.<name>` file, resolving
+/// `${VAR}` references against values defined earlier in the same file.
+fn env_file_mailbox_url(name: &str) -> String {
+    let path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join(format!("../../.env.{name}"));
```
guillemcordoba: Wouldn't it be better to receive MAILBOX_URL here and use it, instead of coupling this process to the names and location of these files?

`ENV=testing just test` will set the MAILBOX_URL to the testing environment automatically.

crates/dashchat-node/src/testing/mod.rs
```diff
@@ -1,11 +1,23 @@
...
 pub use introduce::*;
+pub use mailbox::*;
 pub use test_node::*;
 use tracing_subscriber::EnvFilter;
 
+/// Appends a random suffix to `bytes` so blob content (and therefore its
+/// content-addressed hash) is unique per call. Tests running against a
+/// persistent/shared mailbox must not reuse blob content across runs, or
+/// their blob hashes collide with earlier runs' blobs.
+pub fn unique_blob_bytes(bytes: impl Into<Vec<u8>>) -> Vec<u8> {
```
guillemcordoba, line 15: 🥳

crates/dashchat-node/src/mailbox.rs
```diff
@@ -63,12 +63,58 @@ impl From<MailboxOperation> for Operation {
     }
 }
 
+/// A mailbox server's canonical id (its EndpointId) and dialing address, as
+/// reported by its `/health` endpoint.
+pub struct MailboxHealth {
```
guillemcordoba, line 68: We should move this functionality to the `mailbox-client` crate so it's reused across the project.

e2e-tests/setup/test-env.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function testEnvMailboxUrl(): string | null {
+	const name = process.env.DASHCHAT_TEST_ENV;
+	if (name === undefined || name === '') return null;
+	if (!ALLOWED_TEST_ENVS.includes(name)) {
+		throw new Error(
+			`DASHCHAT_TEST_ENV=${name} is not an allowed test environment (allowed: ${ALLOWED_TEST_ENVS.join(', ')})`,
+		);
+	}
+	const file = path.resolve(__dirname, '..', '..', `.env.${name}`);
```
guillemcordoba: Same here, I would prefer to use MAILBOX_URL instead of reading the .env files.

guillemcordoba (commented):
Looks really good, except the direct reading of the .env files from both rust and typescript tests. I would prefer the tests just receive a MAILBOX_URL that is automatically set by just  (`ENV=testing just test` sets the MAILBOX_URL env variable to whatever is in .env.testing). What do you think @maackle ?

maackle:
@guillemcordoba I applied your feedback about just using MAILBOX_URL. Claude disagrees but I agree with you

PR #420, opened 2026-07-02
E2E tests can disabled P2P syncing in a targetted way. 

jadehopepunk:
Disable it for the waiting-profile tests as the presence of p2p sync has made these tests (which turn off the mailbox to simulate a lack of connectivity) flaky.

### DIFF
--- a/crates/dashchat-node/src/node.rs
+++ b/crates/dashchat-node/src/node.rs
@@ -511,6 +511,17 @@ impl Node {
         Ok(self.endpoint.endpoint().await?)
     }
 
+    /// Tear down all p2p connectivity by closing the iroh endpoint. Afterwards
+    /// the node can neither dial nor accept peer connections, so no gossip sync
+    /// happens; local reads and writes against the op store keep working. This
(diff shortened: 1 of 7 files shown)


crates/dashchat-node/src/node.rs
```diff
@@ -437,6 +437,17 @@ impl Node {
         Ok(self.endpoint.endpoint().await?)
     }
 
+    /// Tear down all p2p connectivity by closing the iroh endpoint. Afterwards
+    /// the node can neither dial nor accept peer connections, so no gossip sync
+    /// happens; local reads and writes against the op store keep working. This
+    /// is a one-way switch intended only for e2e tests that must observe
+    /// pre-sync UI state without racing a direct p2p connection between two
+    /// agents running on the same machine.
+    pub async fn disconnect_p2p(&self) -> Result<()> {
```
maackle: I prefer naming it for what it is to avoid confusion about the implications for mailbox media sync

```suggestion
    #[cfg(feature = "testing")]
    pub async fn close_iroh_endpoint(&self) -> Result<()> {
```

PR #422, opened 2026-07-02
Fix stale mailbox addresses

guillemcordoba:
Fixes the download of media in Android when the sender is not online.

This exposes a security vulnerability where peers can override other peers' addresses in Local Message Servers. I see a couple of possibilities to fix this, but I wanted to talk to @maackle before finishing this PR.

### DIFF
--- a/crates/dashchat-node/src/node/actor.rs
+++ b/crates/dashchat-node/src/node/actor.rs
@@ -289,6 +289,14 @@ impl Actor {
     /// dialable again. Callers only ever pass mailbox `/health` self-addresses
     /// and addresses a user's opt-in local mailbox forwards, so there is no
     /// untrusted address here to guard an existing entry against.
+    //
+    // KNOWN LIMITATION: a malicious client that registered this endpoint
+    // before p2panda discovered it via mDNS/gossip can continue to inject
+    // undialable addresses here (griefing). We cannot detect the upgrade
+    // from mailbox-discovered to node-discovered without a
+    // p2panda discovery hook;
+    // the iroh QUIC handshake prevents data from flowing to the wrong peer,
+    // so the worst case is wasted dial attempts.
     async fn handle_register_peer_addr(
         &mut self,
         addr: iroh::EndpointAddr,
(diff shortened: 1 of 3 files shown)


crates/dashchat-node/src/node/actor.rs
```diff
@@ -286,29 +279,21 @@ impl Actor {
...
+    /// Insert (or refresh) a peer's dialing address in the p2panda address book.
+    ///
+    /// Always overwrites any existing entry: `insert_node_addr` replaces the
+    /// whole `NodeInfo`, resetting its metrics, so an entry marked "stale" by an
+    /// earlier failed dial — which `AddressBookDiscovery` then refuses to
+    /// resolve, leaving the peer undialable and, since the address book is
+    /// persisted, staying that way across restarts — is refreshed and becomes
+    /// dialable again. Callers only ever pass mailbox `/health` self-addresses
+    /// and addresses a user's opt-in local mailbox forwards, so there is no
+    /// untrusted address here to guard an existing entry against.
```
maackle, line 291:
```suggestion
    /// Always overwrites any existing entry: `insert_node_addr` replaces the
    /// whole `NodeInfo`, resetting its metrics, so an entry marked "stale" by an
    /// earlier failed dial — which `AddressBookDiscovery` then refuses to
    /// resolve, leaving the peer undialable and, since the address book is
    /// persisted, staying that way across restarts — is refreshed and becomes
    /// dialable again. Callers only ever pass mailbox `/health` self-addresses
    /// and addresses a user's opt-in local mailbox forwards, so there is no
    /// untrusted address here to guard an existing entry against.
    //
    // KNOWN LIMITATION: a malicious client that registered this endpoint
    // before p2panda discovered it via mDNS/gossip can continue to inject
    // undialable addresses here (griefing). We cannot detect the upgrade
    // from mailbox-discovered to node-discovered without a
    // p2panda discovery hook;
    // the iroh QUIC handshake prevents data from flowing to the wrong peer,
    // so the worst case is wasted dial attempts.
```

maackle (approved):
I think the benefit of fixing the known griefing attack is not worth the benefit of keeping it simple right now, so I think this is good as-is. I added a comment about the griefing.

PR #423, opened 2026-07-02
Fix iOS crashes on app pause

guillemcordoba:
iOS requires that when the app is put on pause and before it's totally closed all handles to files be released. In our case, that means that we need to carefully track and close our tasks when the app is paused.

- Centralize tasks from the tauri app into AppNode for easy stop & cleanup.
- Implement on_pause and on_resume for iOS.


src-tauri/src/app_node.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Cloud-mailbox registration retry, tracked so [`AppNode::pause`] can cancel
+    /// the token and wait for it to drain before the node is torn down. A fresh
+    /// tracker+token pair is installed per node generation in [`AppNode::resume`]
+    /// (`TaskTracker`/`CancellationToken` from `tokio-util`).
+    tracker: TaskTracker,
+    token: CancellationToken,
+    /// Local-mailbox mDNS discovery for the current node generation. Held so it is
+    /// aborted (dropping the handle tears down its browse + interface-watcher
+    /// tasks) when the node is paused, rather than leaking against the dead node.
+    mdns_discovery: Option<AbortOnDropHandle<()>>,
```
maackle, line 25: I think this is OK at this stage but at some point we may want more structured lifecycle management: a task manager that tracks tasks at startup so it knows what to shut down in a general way. It may eventually get too hairy to continue doing it manually. I wrote https://github.com/holochain/task-motel-rs for this purpose years ago, maybe there is something better out there now but at some point maybe we want to include something like that.

maackle (approved):
Looks good

PR #427, opened 2026-07-02
Retry publishing blobs to message server

maackle:
Nodes store which mailbox they've requested to store blob hashes, and detect when that mailbox has actually successfully fetched them. For any unfetched hashes (which are persisted in LocalStore), the node periodically attempts to re-send them to the mailbox, as well as on app startup and (soon) on network change. This ensures media eventually gets to the mailbox and puts the responsibility on the node.


crates/dashchat-node/src/stores/local_store.rs
```diff
@@ -414,6 +419,83 @@ impl LocalStore {
...
+        Ok(())
+    }
+
+    pub async fn remove_unfetched_blobs(
+        &self,
+        mailbox_id: &str,
+        hashes: &[iroh_blobs::Hash],
+    ) -> anyhow::Result<()> {
+        for hash in hashes {
+            self.remove_unfetched_blob(mailbox_id, *hash).await?;
```
guillemcordoba, line 459: Could be more efficient with a single SQL query.

crates/dashchat-node/src/stores/local_store.rs
```diff
@@ -414,6 +419,83 @@ impl LocalStore {
...
+    }
+
+    /// Remove every `unfetched_blob_hashes` row for these hashes across ALL
+    /// mailboxes. Called when a blob is deleted locally, so the followup task
+    /// stops re-announcing a hash this node can no longer serve.
+    pub async fn remove_unfetched_blobs_all_mailboxes(
+        &self,
+        hashes: &[iroh_blobs::Hash],
+    ) -> anyhow::Result<()> {
+        for hash in hashes {
```
guillemcordoba, line 471: Again, can be more efficient by using a single SQL query.

guillemcordoba (approved):
LGTM. 

I do think in the near future we want to take a step back and reconsider the state of the codebase regarding the approach we have taken with supporting media and the usage of iroh blobs. I'd like to be able to simplify, or at least, structure the codebase in a way that encapsulates each concern into its own separate module.

PR #428, opened 2026-07-03
Reduce shared contact code length

jadehopepunk:
Goal: Shorten the QR Code.

Result: This PR drops the QR code from 348 chars to 108 chars. Further optimisations are possible in future PRs

This is done by:

* Expressing time in hours since Epoch
* Ensuring that we don't convert to hex before converting to base64
* Removing agent_id from the contact code

In hindsight, these should have been separate PRs because this has gotten quite large. The bulk of this is the removal of agent ids from the contact code. To achieve this, when someone scans your QR code and requests to be contacts, they also send a topic ID that you can send an acceptance message back on, which includes your agent id.

### DIFF
--- a/crates/dashchat-node/src/contact.rs
+++ b/crates/dashchat-node/src/contact.rs
@@ -48,9 +48,28 @@ pub enum ShareIntent {
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct InboxTopic {
     // NOTE: order of these fields matters! expires_at, then topic.
-    /// Expiry date must be within the valid range expressible by DateTime::from_timestamp_nanos
+    /// Expiry date. Serialized as a whole number of hours since the Unix epoch
+    /// to keep the QR code short; sub-hour precision is not needed for expiry.
+    #[serde(with = "expires_at_hours")]
     pub expires_at: DateTime<Utc>,
     pub topic: Topic<kind::Inbox>,
+}
+
+/// Serialize a `DateTime<Utc>` as an `i64` count of whole hours since the Unix
+/// epoch. Truncates toward the epoch; used only for coarse expiry timestamps.
+mod expires_at_hours {
+    use chrono::{DateTime, Utc};
+    use serde::{Deserialize, Deserializer, Serializer};
+
+    pub fn serialize<S: Serializer>(dt: &DateTime<Utc>, s: S) -> Result<S::Ok, S::Error> {
+        s.serialize_i64(dt.timestamp().div_euclid(3600))
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(d: D) -> Result<DateTime<Utc>, D::Error> {
+        let hours = i64::deserialize(d)?;
+        DateTime::from_timestamp(hours * 3600, 0)
+            .ok_or_else(|| serde::de::Error::custom("expires_at hours out of range"))
+    }
 }
 
 impl std::fmt::Display for QrCode {
@@ -109,7 +128,9 @@ mod tests {
             device_pubkey: DeviceId::from(pubkey),
             inbox_topic: Some(InboxTopic {
                 topic: Topic::inbox(),
-                expires_at: Utc::now() + chrono::Duration::seconds(3600),
+                // Hour-aligned so it survives the coarse (hours-since-epoch)
+                // serialization used to keep the QR code short.
+                expires_at: DateTime::from_timestamp(1_700_000_000 / 3600 * 3600, 0).unwrap(),
             }),
             agent_id,
             share_intent: ShareIntent::AddDevice,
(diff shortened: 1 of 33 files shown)


crates/dashchat-node/src/contact.rs
```diff
@@ -37,29 +36,47 @@ pub struct QrCode {
...
+    /// Expiry date. Serialized as a whole number of hours since the Unix epoch
+    /// to keep the QR code short; sub-hour precision is not needed for expiry.
+    #[serde(with = "expires_at_hours")]
     pub expires_at: DateTime<Utc>,
     pub topic: Topic<kind::Inbox>,
 }
 
+/// Serialize a `DateTime<Utc>` as an `i64` count of whole hours since the Unix
+/// epoch. Truncates toward the epoch; used only for coarse expiry timestamps.
+mod expires_at_hours {
```
maackle: Hours may be coarse: the typical in-person QR code exchange should only last minutes. I would think a typical expiry time would be 10 minutes (similar to an SMS 2FA code). Maybe change this to minutes instead of hours? That's still well within i64. Btw, why not u64?

    (2026−1970)×365×24×60 = 29_433_600

crates/dashchat-node/src/node.rs
```diff
@@ -434,11 +434,26 @@ impl Node {
...
+    /// Resolve the agent id we've recorded for a device pubkey, if any. Returns
+    /// `None` until the contact is established (e.g. while an outgoing contact
+    /// request is still pending its ack).
+    pub async fn agent_for_device(
+        &self,
+        device_pubkey: DeviceId,
+    ) -> Result<Option<AgentId>, Error> {
+        let map = self
+            .local_store
+            .lookup_contacts([&device_pubkey])
```
maackle: could just as well use `lookup_contact_by_device_id` for a single lookup

crates/dashchat-node/src/node.rs
```diff
@@ -1237,55 +1266,217 @@ impl Node {
...
+
+    /// Returns true if we have an outgoing contact request recorded for
+    /// `device_pubkey` (i.e. we scanned their code and are awaiting their ack).
+    pub(crate) async fn has_outgoing_pending_request(
+        &self,
+        device_pubkey: DeviceId,
+    ) -> anyhow::Result<bool> {
+        let found = self
+            .op_store
+            .get_interleaved_logs(self.device_group_topic().into(), vec![self.device_id()])
```
maackle: ack, this is my bad -- this is behind the "testing" feature, but I had that as a default feature. The intention is that this should not be used in production, because it's inefficient. I'll make a PR to take out that default feature and you can rewrite this in terms of a different query -- maybe even one exactly like this one, but I just want to keep this function away exactly because it's so convenient to grab in a case like this.

crates/dashchat-node/src/node/app_processing.rs
```diff
@@ -428,17 +428,98 @@ impl Node {
...
+                    .get_reply_inbox_topics()
+                    .await?
                     .iter()
-                    .any(|it| *it.topic == TopicId::from(topic))
-                {
-                    // not for me, ignore
+                    .any(|it| *it.topic == topic_id);
+                if !is_advertised_topic && !is_reply {
+                    // not for me (e.g. another scanner's request on a shared
+                    // advertised inbox we only transiently synced): ignore.
```
maackle:
```suggestion
                    // advertised inbox we only synced as an intermediary): ignore.
```

crates/dashchat-node/src/node/app_processing.rs
```diff
@@ -428,17 +428,98 @@ impl Node {
...
+                                    }
+                                });
+                            }
+                        }
+                    }
+                    InboxPayload::ContactRequestAck { profile, agent_id } => {
+                        // The op must arrive on our private reply topic and be
+                        // signed by the device whose QR we scanned. Verifying
+                        // Verifying author == expected_ack_author prevents a
+                        // third party that transiently synced the advertised
```
maackle: there's that "transiently" again, I suggested "synced as an intermediary" earlier but is that what you meant? I'm not quite sure what's meant by "transiently"

crates/dashchat-node/src/testing/behavior.rs
```diff
@@ -31,30 +31,42 @@ impl Behavior {
...
+        // ack (it learns our agent_id only then), so wait for it to land before
+        // returning a fully-established mutual contact.
+        let me = self.node.agent_id();
+        let deadline = std::time::Instant::now() + Duration::from_secs(15);
+        while !other.get_contacts().await?.contains(&me) {
+            if std::time::Instant::now() >= deadline {
+                anyhow::bail!("scanner did not record the contact in time");
+            }
+            tokio::time::sleep(Duration::from_millis(100)).await;
+        }
```
maackle: FYI this polling pattern is encapsulated by https://github.com/dash-chat/dash-chat/blob/e79c6fcf11471b10a3d63bcf2c322609b056dd03/crates/dashchat-node/src/testing/test_node.rs#L267 if you want to reuse it and increase the chances that it gets reused

maackle (changes requested):
The flow looks good overall, the complexity seems necessary to accomplish what needs to be accomplished, and it goes beyond just a smaller QR code to add a real security gain.

I need to get a PR in in front of this to turn off the testing feature flag, and some minor changes requested

maackle (approved):

maackle:
Claude makes some good points, I'd say address those first then merge

PR #430, opened 2026-07-03
Fix iOS notifications

guillemcordoba:
`iroh-blobs` requires a single process is using a data path with a file lock. The iOS Push Notification Extension runs in a separate process, but it needs to run a `dashchat-node` to be able to ingest operations coming from the mailbox.

The solution is a "no blobs" option for `dashchat-node`.


maackle (approved):
Looks good. I hope eventually we get to do a big pass over all this and refactor so it all makes more sense and there's less to keep in our heads (not just your PRs but mine especially!)

PR #432, opened 2026-07-03
Adds a .env.local which allows all rust tests to run using a local toy mailbox instead of MemMailbox

maackle:


crates/dashchat-node/src/mailbox.rs
```diff
@@ -88,20 +88,20 @@ pub async fn fetch_mailbox_health(base_url: &str) -> anyhow::Result<MailboxHealt
...
-        .post(&url)
-        .json(&mailbox_client::RegisterPeerRequest { addr: our_addr })
-        .send()
-        .await?
-        .error_for_status()?;
-    Ok(())
+impl crate::Node {
+    /// POST our current `EndpointAddr` to a mailbox's `/peers/register` endpoint
+    /// so it can dial us when fetching blobs we published.
+    pub async fn register_with_mailbox(&self, base_url: &str) -> anyhow::Result<()> {
```
guillemcordoba, line 94: Hum... I like this. I think it's good to merge for now.

Although I think this should probably be handled at the mailbox client level. The `Mailboxes` manager should probably deal with all that's required to talk and sync to a mailbox server.

.env.local
```diff
@@ -0,0 +1 @@
+DASHCHAT_SPAWN_LOCAL_MAILBOX=true
```
guillemcordoba, line 1: I wonder about this... This file introduces a totally different concept of an "environment". In the other .env files, we set URLs to servers. Here, we are configuring how we want local tests to run. I don't think we'd want to mix those concepts up... 

A question comes up. If we are testing with a real local mailbox server, do we need to test with the memory one?

If so, can we have another pattern to configure how the tests would run? Not sure the best one, let's discuss.

guillemcordoba (commented):
Overall I really like this change, a step in the right direction. I think it's good to merge except the way to configure the tests running with a local mailbox.

guillemcordoba:
Will merge for now, changing how this test mode is activated can be a follow up if needed.

PR #435, opened 2026-07-06
Try uploading blobs to mailbox immediately on publish

maackle:
Adds a `/blobs/upload` endpoint which is used during payload publishing during log sync.
Instead of just asking the server to store blob hashes that it will fetch "later" (now renamed to `/blobs/register-hashes`), clients now register hashes as before, but also immediately upload blobs. 
This upload is a best-effort attempt to get the blobs to the server as quickly as possible.

The server will remove from its pending fetch pool the hashes of any blobs received via upload. Also, now when the server registers hashes, it includes a grace period during which the server will not attempt to fetch those blobs, which gives the client time to upload and prevents duplicates. Whenever an upload is successful, it refreshes the grace period for all other hash entries whose grace period has not yet expired.

### DIFF
--- a/crates/mailbox-client/src/toy.rs
+++ b/crates/mailbox-client/src/toy.rs
@@ -95,24 +95,28 @@ impl<Item: MailboxItem> ToyMailboxClient<Item> {
         }
     }
 
-    /// Attach a blob-bytes source so `publish` pushes blob bytes straight to the
-    /// mailbox (best-effort) before announcing their hashes. Without a reader the
-    /// client only announces hashes and the mailbox fetches the bytes from us.
+    /// Attach a blob-bytes source so `publish` streams blob bytes to the mailbox
+    /// (best-effort, in a spawned task) after announcing their hashes. Without a
+    /// reader the client only announces hashes and the mailbox fetches the bytes
+    /// from us.
     pub fn with_blob_reader(mut self, blob_reader: std::sync::Arc<dyn crate::BlobReader>) -> Self {
         self.blob_reader = Some(blob_reader);
         self
     }
 
-    /// Push blob bytes to the mailbox immediately (best-effort), then announce
-    /// their hashes and reconcile the unfetched tracker: record hashes the
-    /// mailbox still needs, remove any it already has. The announce always runs,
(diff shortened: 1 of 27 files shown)


crates/mailbox-client/src/toy.rs
```diff
@@ -41,17 +76,39 @@ pub async fn send_store_blobs(
...
+}
+
 /// A client for the toy mailbox server.
 #[derive(Clone)]
 pub struct ToyMailboxClient<Item: MailboxItem> {
     id: MailboxId,
     base_url: String,
     sender_pubkey: iroh::EndpointId,
     tracker: std::sync::Arc<dyn crate::UnfetchedBlobTracker>,
+    blob_reader: Option<std::sync::Arc<dyn crate::BlobReader>>,
```
guillemcordoba, line 111: Why is blob_reader optional? It seems to me like all mailbox clients need to be able to read the blobs.

guillemcordoba, line 111: Also, why do we need the trait? Can't we extract the type into a separate crate that is used by both `dashchat-node` and `mailbox-client`? I'd like to reduce the complexity in terms of traits and optional arguments as much as possible.

crates/mailbox-client/src/toy.rs
```diff
@@ -365,14 +491,148 @@ mod tests {
         });
         let base_url = format!("http://{addr}");
 
-        let tracker = std::sync::Arc::new(RecordingTracker::default());
-        let already = crate::toy::send_store_blobs(
+        let _tracker = std::sync::Arc::new(RecordingTracker::default());
```
guillemcordoba: Remove this variable if it's not needed anymore.

guillemcordoba (approved):
Merging for now as this should improve the media upload UX.

Overall, I'm not quite happy with the state of media handling in the backend of the app. There is a lot of complexity in both the mailbox client and server, specially with the round trip way to register hashes to the server to have the server pull from the clients, and the fetch pool / blob sync logic. There are also privacy implications which may cause issues in the future.

I think in the near future we should have a session where we draw out the current architecture of all the components, and discuss possible simplifications to it.

PR #436, opened 2026-07-06
DRY up explicit local mailbox pattern into function

maackle:


maackle:
This is just cleanup so I'm merging without review

PR #437, opened 2026-07-07
Edit messages frontend

maackle:
Implement the initial pass for the edit messages UI. Includes the spotlight overlay menu that signal has on mobile when longpressing a message. Inspecting message edit history is out of scope for this PR.


packages/stores/src/chats/edits.ts
```diff
@@ -0,0 +1,143 @@
...
+			current = op.kind.target;
+			continue;
+		}
+		return undefined;
+	}
+	return undefined;
+}
+
+/** Whether an edit op may be applied to its target. Mirrors `validate_edit`. */
+function isValidEdit(
```
jadehopepunk: This seems pretty dangerous to mimic the backend code on the frontend. It's easy to end up with differences between the. 

Obviously this is something that happens with form validation and so on. If we need to do it passively (like knowing which messages we can display an edit button next to) that might be necessary.

But, since the "backend" is in fact on the same device, can't we just throw to the backend to answer if it's valid?

ui/src/lib/components/messages/composer/MessageComposer.svelte
```diff
@@ -149,12 +183,36 @@
 		class="message-input-bar"
 		class:pb-safe={!showMediaPanel && !keyboard.isOpen}
 	>
-		{#if !isMobile}
+		{#if editing}
```
jadehopepunk: MessageComposer is a complex beast, and may get more so. I'd suggest extracting this new edit code as a component.

ui/src/lib/components/messages/EditedIndicator.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
```
jadehopepunk, line 1: I love this component, small and one-use :)

ui/src/lib/components/messages/EditHistorySheet.svelte
```diff
@@ -0,0 +1,61 @@
...
+	</div>
+	<Block>
+		<h2 class="mb-3 text-lg font-semibold">{m.editMessageHistory()}</h2>
+		<div class="flex flex-col gap-3">
+			{#each versions as version, i (i)}
+				<div class="flex flex-col gap-1">
+					<div class="quiet flex items-center gap-2 text-xs">
+						<MessageTimestamp timestamp={version.timestamp} class="quiet" />
+						{#if i === versions.length - 1}
+							<span>· {m.originalMessage()}</span>
```
jadehopepunk: What is this middle-dot character? How will that be read by a screen reader.

Also, it's in a flex row, where it goes timestamp {gap} middle dot {space} original message. Perhaps it should be in it's own span so that it has two equal gaps.  That span could have aria-hidden.

Alternatively, since it's more like display decoration, it could be moved into the CSS.

ui/src/lib/components/messages/message-editing.svelte.ts
```diff
@@ -0,0 +1,24 @@
+import type { Message, MessagesStore } from 'dash-chat-stores';
```
jadehopepunk: I don't think this is in the right place. It's not a component.

jadehopepunk: Hmm, I see we have at least one other example of helper functions in this dir. So, not inconsistent, but I tend to keep my "components" folder only for components, and create a src/lib/helpers/messages or something for the rest

jadehopepunk (commented):
Not explicitly approved. I'd be keen to hear if you think we can move the validation logic just to the back end, to avoid the duplication. 

And of the various comments on the front end code, there's one small component extraction that I think is important.

ui/src/lib/components/messages/QuickReactionBar.svelte
```diff
@@ -87,6 +102,15 @@
 				{emoji}
 			</button>
 		{/each}
+		{#if canEdit}
```
guillemcordoba: The edit button shouldn't be in the QuickReactionBar.

I think we should create an additional element called MessageActions, that both has the QuickReactionBar above the element, and the message actions menu below it.

guillemcordoba:
@maackle could you give this a review?

ui/src/lib/components/messages/composer/DiscardEditButton.svelte
```diff
@@ -0,0 +1,29 @@
...
+	{onClick}
+	label={m.cancel()}
+	testid="composer-cancel-edit"
+	class={isWideScreen.value
+		? '!bg-[#848484] !text-white !opacity-100'
+		: `!h-[42px] !w-[42px] !opacity-100 ${
+				theme === 'ios'
+					? '!bg-ios-light-glass shadow-ios-light-glass backdrop-blur-lg dark:!bg-ios-dark-glass dark:shadow-ios-dark-glass'
+					: '!bg-[#545863] !text-white dark:!bg-[#BEBFC5] dark:!text-[#1E2438]'
+			}`}
```
maackle, line 28: this style seems overly specific, are there some variables to tap into?

ui/src/lib/components/messages/composer/MessageComposer.svelte
```diff
@@ -61,14 +64,63 @@
 	let sending = false;
 
 	let showMediaPanel = $state(false);
-	// Stays true through the panel→keyboard swap so the panel remains visible
-	// under the rising keyboard; cleared by the slot's onHidden once covered.
-	let mediaPanelMounted = $state(false);
+
+	let editing = $state<Message | null>(null);
+	/** Edit requested while a draft was present, awaiting discard confirmation. */
+	let pendingEdit = $state<Message | null>(null);
```
maackle: I would somewhat prefer consolidating both of these into one type but what's here is also OK

```ts
let editing = $state<{message: Message, pending: bool} | null>(null);
```

ui/src/lib/components/BelowKeyboardSurface.svelte
```diff
@@ -0,0 +1,56 @@
+<!--
+	Render content below the keyboard — the counterpart to `renderAboveKeyboard`.
+	The bits above the keyboard (composer bar, message list) glide up; this
+	surface is revealed in the freed space, standing in for the keyboard. The
+	children stay mounted while the plugin reports the surface visible, so they
+	don't vanish mid-swap while the rising keyboard is still covering them.
+
+	Pins itself to the viewport bottom; it must not sit inside a
+	`renderAboveKeyboard` node — a registered ancestor's transform would drag it
+	along instead of leaving it pinned, and would break `position: fixed`.
```
maackle: I think renderAboveKeyboard was removed?

maackle (commented):
Looks good! I started the PR so I can't approve, but I do.

ui/src/lib/components/messages/composer/DiscardEditButton.svelte
guillemcordoba: Agreed. However, I'm going to leave as is, and create a follow up PR in which an IconButton "appearance" prop is introduced, that is a variant of all the different ways our IconButtons can look in the app.

ui/src/lib/components/BelowKeyboardSurface.svelte
guillemcordoba: renderBelowKeyboard was the one that got removed, replaced by this element actually.

PR #439, opened 2026-07-08
Delete message

maackle:
## Followup

- [ ] find places where eager sync might propagate a deleted message when being a little more careful would have allowed a tombstone to delete it before syncing
- [x] ~~restructure the mailbox to allow deleting payloads independently of headers~~ we decided to punt on this until the next iteration of mailbox design

(description mostly copied from #438)

## Summary

Adds support for deleting messages for everyone in both direct and group chats. When a user deletes a message, the entire edit chain is tombstoned and payloads are dropped. Subsequent edits to the message are invalid.

## Notable Implementation Details

- **Mailbox safety**: Operations are only transmitted to mailboxes once they've completed app-layer processing. This prevents leaking payloads of operations that are about to be tombstoned by a delete.
- **Complete chain deletion**: A delete must cover the entire edit chain (original + all edits by any author), preventing partial deletions that would leave orphaned edits.
- **Competing deletes**: Resolved like competing edits—earliest (by seq_num, hash as tiebreaker) wins; later deletes covering the same ops are rejected.
- **New member protection**: Members added after a delete only receive body-less operation headers, never the deleted payloads or media.
- **Blob cleanup**: Deleted media is unpinned from the blob store and removed from fetch tracking, making it eligible for garbage collection.


crates/dashchat-node/src/node/app_processing.rs
```diff
@@ -242,7 +261,52 @@ impl Node {
...
+            self.local_store.add_tombstone(topic, *hash).await?;
+            if op.body.is_some() {
+                self.unprocess_app(&op).await?;
+                self.op_store.delete_body(hash).await?;
+            }
+            scrubbed.push(MailboxOperation {
+                topic,
+                header: op.header,
+                body: None,
+            });
```
maackle: I'm either going to remove the scrubbing feature for now, or give a little extra attention to the mailbox server to "ensure" (in a currently unauthenticated way) that only the author can scrub data, and that data can never be "unscrubbed" once scrubbed.

packages/stores/src/chats/deletes.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Every message covered by a delete is replaced with a "deleted" placeholder
+ * anchored at the chain's root operation — regardless of whether the deleted
+ * payloads are still present locally: the backend tombstones them, and a
+ * member that joined after the delete only ever sees body-less operations. A
+ * delete only counts when authored by the same device as the operations it
+ * covers, mirroring the backend's authorship rule. */
+export function applyDeletes(
+	messages: Record<Hash, Message>,
+	logs: Record<DeviceId, SimplifiedOperation<Payload>[]>,
+): void {
```
maackle: @guillemcordoba and @jadehopepunk can you recommend an alternative to duplication of logic in the front end? This is a front end architectural decision I don't want to make.

crates/dashchat-node/src/node/app_processing.rs
maackle: I removed scrubbing, leaving it for a followup

packages/stores/src/chats/deletes.ts
jadehopepunk: @maackle yes I agree with don't want that duplication. 

The front end is absolutely intended to be a local cache of data that isn't canonical. So, this is kind of a cache invalidation problem.

I think that for deletes we should go with a blunt instrument. When we receive notify the frontend of a delete operation, we invalidate on a large scale, rather than go through and apply the delete (and possibly get a different answer from the backend).

Can we start with invalidate the entire chat if we receive a delete operation for a message, or the entire contacts store if we receive a delete operation for a contact?

Invalidation in this sense means re-fetching all the store data from the backend. So we rely on the backend to perform the more careful deletion.

guillemcordoba: This is how I would do it: https://github.com/dash-chat/dash-chat/pull/448

crates/dashchat-node/src/chat/validation/delete.rs
```diff
@@ -0,0 +1,527 @@
...
+    ///
+    /// Rules:
+    /// - every hash must resolve to a `Message` or `EditMessage` operation,
+    /// - the set must be exactly one complete edit chain (one original message,
+    ///   every edit of any member included, no gaps, no extras),
+    /// - no other delete may already cover any of the hashes,
+    /// - the deleter must be the author of every operation in the set,
+    /// - the delete must fall within [`DELETE_WINDOW_MICROS`] of the original
+    ///   message.
+    pub fn validate(&self, valid_ops: &ValidChatOps) -> Result<(), DeleteError> {
```
maackle, line 111: Naming: this is only part of the full validation, which includes authorship check

crates/dashchat-node/src/node/app_processing.rs
```diff
@@ -206,6 +206,15 @@ impl Node {
...
                                 }
 
+                                // Acknowledge the operation now that application-layer
+                                // processing has finished. The node uses an `Explicit`
+                                // ack policy, so this persisted ack is what makes the
+                                // operation eligible for mailbox transmission (see
+                                // `OpStore::acked_log_height`).
+                                if let Err(err) = operation.ack().await {
+                                    tracing::error!(?err, "failed to acknowledge operation");
+                                }
```
maackle:
```suggestion
```

crates/dashchat-node/src/node/app_processing.rs
maackle, line 367: missing acknowledgement??

guillemcordoba:
@maackle in [this comment](https://github.com/dash-chat/dash-chat/pull/439#issuecomment-5117444662), Claude found a security issue, which I reproduced and created a test for [here](https://github.com/dash-chat/dash-chat/pull/439/changes/7145701d2895a9d13e6f7a0d22e7c38424fe795f). 

I found another security issue in [this line](https://github.com/dash-chat/dash-chat/pull/439/changes#diff-92a891cd0f68d1e72baf485b85a68a5ba24116b0f65c4a3a90d34498e97a9bbbR367), for which I created [another test](https://github.com/dash-chat/dash-chat/pull/439/changes/8f356fb146d712a52267c56b2ddf6bee288b48c2).

I think we should fix the behavior now rather than be surprised by a malfunctioning of the app or a failed security review in the future. Other than fixing the behavior of the validate functionality to make the tests pass, I think the PR is good to go.

guillemcordoba (approved):

PR #440, opened 2026-07-09
Add delete-for-me message deletion feature

maackle:
Implements "delete for me" functionality, allowing users to delete any message (sent or received) from their own devices only, without affecting other participants. This mirrors Signal's UX where delete-for-me removes messages with no placeholder, in contrast to delete-for-everyone which leaves a tombstone.

This uses a lot of the same tombstoning machinery as "delete for everyone". Also added in this PR is a `TombstoneReason` which distinguishes between "for me" and "for everyone". This allows "delete for me" to be transitively applied to all future EditMessage operations which may come in, even if previous edits were already tombstoned. This is important because unlike "delete for everyone", "delete for me" can apply to messages not authored by me.

## Implementation Details

- Delete-for-me operations live in the author's private device group topic, ensuring only their own devices see the deletion
- Only the original message is referenced in DeleteForMe, with the effect that any EditMessage part of the same edit chain as the original message also be considered "deleted for me" (payload dropped and tombstoned ASAP).
- Unlike delete-for-everyone, there is no delete window or authorship restriction—any message can be deleted from one's own devices
- Messages deleted for-me are completely removed from the UI (no placeholder), following Signal's UX
- The shared chat mailbox and peer copies remain untouched; the message stays visible to other participants
- Tombstones are recorded against the chat topic, so incoming deleted operations are dropped during processing


crates/dashchat-node/src/stores/op_projection.rs
```diff
@@ -38,10 +39,39 @@ const MIGRATIONS: &[&str] = &[
...
+    }
+
+    fn from_db(value: &str) -> anyhow::Result<Self> {
+        match value {
+            "for_everyone" => Ok(TombstoneReason::DeletedForEveryone),
+            "for_me" => Ok(TombstoneReason::DeletedForMe),
+            other => Err(anyhow::anyhow!("unknown tombstone reason: {other}")),
+        }
+    }
+}
```
maackle, line 78: Use ints or at least better strings

packages/stores/src/chats/messages-store.ts
```diff
@@ -73,38 +74,25 @@ export class MessagesStore {
...
-				}
-			}
-		}
-
-		return hashes;
+		if (chatId === '') return [];
+		// Establish reactive dependencies on the logs so this recomputes when a
+		// delete or a tombstoned edit is processed into either topic.
+		await this.logsStore.logsForAllAuthors(chatId);
+		await this.contactsStore.devicesStore.myDeviceGroupTopic();
```
maackle: @guillemcordoba is this canonical signalium use or does it smell?

packages/stores/src/mock/messages-client.ts
```diff
@@ -78,10 +78,80 @@ export class MockMessagesClient implements IMessagesClient {
 	}
 
 	async deleteMessageForMe(chatId: ChatId, targetHash: Hash): Promise<Hash> {
-		const hashes = await this.editChainHashes(chatId, targetHash);
+		// The store passes the original message hash, but resolve the root
+		// defensively (editChainHashes walks back to it) so the whole message is
+		// named regardless of which version was pointed at.
```
maackle:
```suggestion
		// defensively (editChainHashes walks back to it) so the whole message is
		// referenced regardless of which version was pointed at.
```

crates/dashchat-node/src/stores/op_projection.rs
```diff
@@ -186,6 +216,45 @@ impl OpProjection {
...
+                .await?;
+        row.map(|(reason,)| TombstoneReason::from_db(&reason))
+            .transpose()
+    }
+
+    /// Every tombstone in `topic`, paired with its reason. The frontend uses
+    /// this to drop delete-for-me messages (and their edits) from view while
+    /// keeping the delete-for-everyone placeholders.
+    pub async fn tombstones(
+        &self,
```
maackle: @guillemcordoba do you think it would be better for the frontend to get all tombstones up front like this, or to query the backend for each deleted message it needs to check? (or perhaps query once just for the set of message hashes it needs?)

packages/stores/src/chats/messages-store.ts
guillemcordoba: It's not canonical signalium no... I think the way to do this would be have a tombstone-store, that listens to tomstone-added events from the backend, and has a reactive with all the ids of the tomstoned operations.

packages/stores/src/chats/messages-store.ts
```diff
@@ -161,13 +180,43 @@ export class MessagesStore {
...
+ *
+ * A `DeletedForMe` op (and any of its edits) is removed outright — following
+ * Signal, a delete-for-me leaves no placeholder. A `DeletedForEveryone` edit
+ * that slipped through as a raw body-less placeholder (rather than the
+ * chain-collapsed `'deleted-for-everyone'` root that `applyDeletes` produces
+ * from the `DeleteMessage` op) is removed too, so only the root placeholder
+ * remains. Body-less ops that are *not* tombstoned stay as `'body-unavailable'`
+ * — a genuinely unfetched message, which the tombstone set lets us tell apart
+ * from a deletion. */
+function applyTombstones(
```
guillemcordoba: I don't like functions of this style when reducing style, they mutate parameters so the caller loses visibility on what's happening to the overall state object it's building.

crates/dashchat-node/src/stores/op_projection.rs
guillemcordoba: This is one of the cases in which going from the frontend backwards feels much better. I would take the approach of designing the `TombstoneStore` first, which needs to both load the current tombstones upfront and then be reactive to new to tombstones being created, and then adapt the backend to whatever functions that store needs.

ui/src/lib/components/messages/DeleteMessageDialog.svelte
```diff
@@ -2,29 +2,89 @@
...
+
+	const store: MessagesStore = getContext('messages-store');
+
+	// Delete-for-me is offered on every message; delete-for-everyone only on my
+	// own, and only within the delete window.
+	const forEveryone = $derived(
+		canDeleteMessageForEveryone(message, myDeviceId),
+	);
+
+	async function deleteForEveryone() {
```
guillemcordoba, line 32: I really like the direction of this refactor, I think moving these functions into the element makes the code simpler and more encapsulated. On this note, it may be worth to move the location in which the DeleteMessageDialog is used to MessageActionsMenu directly.

src-tauri/src/commands/chats.rs
```diff
@@ -106,12 +107,49 @@ pub async fn delete_message(
...
+    let header = node
+        .delete_message_for_me(chat_id, target_hash)
+        .await
+        .map_err(|err| format!("{err:?}"))?;
+    Ok(header.hash())
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct Tombstone {
```
guillemcordoba, line 132: This is surprising to see here. Is the reason this other Tombstone struct is here to serialize with camel case? We are not serializing with camelCase in any other struct in the app, we are using snake case in the frontend.

guillemcordoba (changes requested):

crates/dashchat-node/src/stores/op_projection.rs
```diff
@@ -38,10 +39,42 @@ const MIGRATIONS: &[&str] = &[
...
+// Possible solutions include:
+// - Storing EditMessage references on the Header as a custom extension (requires p2panda support)
+// - Persisting either edit chains or tombstones in a non-purgable store so that it can be used for projection reconstruction
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum TombstoneReason {
+    DeletedForEveryone,
+    DeletedForMe,
+}
+
+impl TombstoneReason {
```
guillemcordoba, line 65: Isn't there a more ergonomic way to serialize/deserialize a string in database to an enum using sqlx?

e2e-tests/specs/groups/delete-group-messages.spec.ts
```diff
@@ -25,21 +25,58 @@ describe('Deleting group messages', () => {
...
 
 		await mine.waitForDeleted(await agent1.tr('youDeletedThisMessage'));
 		await theirs.waitForDeleted(
 			await agent2.tr('someoneDeletedThisMessage', { name: 'Alice Test' }),
 		);
 	});
 
-	it('does not offer Delete on another member’s messages', async () => {
-		await agent2.groupChatPage.composer.sendMessage("Bob's message stays");
+	// The delete-for-me op lives in the deleter's private device-group topic and
```
guillemcordoba, line 36: I don't like this comment, the e2e test should not explain the internal logic of the feature.

packages/stores/src/types.ts
```diff
@@ -176,6 +176,24 @@ export interface ReadMessagesPayload {
 	message_hashes: Hash[];
 }
 
+export interface DeleteForMePayload {
+	chat_id: ChatId;
+	/** The original message being deleted. Its edit chain (present and future)
```
guillemcordoba, line 181: Again, I don't think the frontend should explain how the feature works in the backend.

packages/stores/src/types.ts
```diff
@@ -176,6 +176,24 @@ export interface ReadMessagesPayload {
...
+	 * is tombstoned transitively by the backend, so only the root is named
+	 * here — unlike `DeleteMessagePayload`, which lists the whole chain. */
+	message_hash: Hash;
+}
+
+/** Why an operation was tombstoned, mirroring the backend `TombstoneReason`. A
```
guillemcordoba, line 187: I think this comment is redundant

ui/src/lib/components/messages/MessageActionsMenu.svelte
```diff
@@ -3,17 +3,17 @@
...
-	} from './message-helpers';
+	import { canEditMessage } from './message-helpers';
 	import ListAction from '$lib/components/navigation/ListAction.svelte';
 
 	interface Props {
 		message: Message;
 		myDeviceId: DeviceId;
 		onEdit?: () => void;
 		onCopy: () => void;
+		/** Opens the delete dialog. Offered on every message, since delete-for-me
```
guillemcordoba: I think this comment is redundant as well.

guillemcordoba (commented):
I added some more comments. If the two PRs get merged and the comments addressed, I approve this PR!

src-tauri/src/commands/chats.rs
maackle: It's following the pattern of GroupMember in the same file which is serialized with camelCase. Should we change this too? https://github.com/dash-chat/dash-chat/blob/b15bbf4c94778097e0e1d741ebbb6a336f7d44b1/src-tauri/src/commands/chats.rs#L12-L18

maackle: Also there is no other type representing a single tombstone. There are only rows in the database and the new SystemEvent::Tombstones which represents multiple tombstones.

ui/src/lib/components/messages/DeleteMessageDialog.svelte
maackle: Nice idea, that simplifies things

src-tauri/src/commands/chats.rs
guillemcordoba: Oh sorry, yes we should probably change it as well, go ahead with it if you don't think it's too much of a scope creep.

maackle:
@guillemcordoba please give special review to these commits 
image

guillemcordoba (approved):
LGTM. The use:portal action I would like to do away with, somehow. For now I think it would be sufficient to move it to LazyDialog, so that all the dialogs get the same functionality without needing to embed them in divs.

maackle:
@guillemcordoba for now I just reverted the change I made that you suggested, it doesn't seem worth the effort right now but if we have other similar dialogs maybe we can pull that into a nice component that breaks out of its bounding box in a more palatable way. So it's back to DeleteMessageDialog appearing in both MessageFromMe and FromOthers.

PR #442, opened 2026-07-10
Shorten qr code 2

jadehopepunk:
Further shortening of the QR Code

This is done by replacing the 32 byte topic with an 8 byte nonce that can be used to derive the topic. 
And, by switching to an unpadded code to avoid the couple of "==" on the end.

### DIFF
--- a/crates/dashchat-node/src/contact.rs
+++ b/crates/dashchat-node/src/contact.rs
@@ -156,7 +156,7 @@ impl From<QrCode> for String {
 impl TryFrom<String> for QrCode {
     type Error = anyhow::Error;
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        Ok(QrCode::from_str(&value).unwrap())
+        Ok(QrCode::from_str(&value)?)
     }
 }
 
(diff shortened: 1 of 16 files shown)


jadehopepunk:
This is ready for review. Please note that the issue of both typescript and rust handling serialisation/deserialisation of QR codes is bugging me and i'm preparing another PR for that.

crates/dashchat-node/src/contact.rs
```diff
@@ -6,34 +7,59 @@ use std::str::FromStr;
 
 use crate::{DeviceId, Topic, topic::kind};
 
+/// An 8-byte nonce serialized as a hex string at the JSON/Tauri boundary.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize)]
+#[serde(transparent)]
+pub struct InboxNonce(#[serde(with = "hex::serde")] pub [u8; 8]);
```
maackle, line 13: Kind of a shame to reduce a field from 32 bytes down to 8, which then get serialized with 16 bytes.

Maybe use what p2panda's VerifyingKey uses? https://github.com/p2panda/p2panda/blob/bb62866f7012252a74b88dbc11fa77328ad35973/p2panda-core/src/serde.rs#L22

maackle (approved):
Looks great except the serialization recommendation

PR #444, opened 2026-07-11
ACID review

maackle:
- [x] separate state for user-initiated actions vs reduced state from op processing
- [x] use explicit acks for processed ops
- [x] better documentation for what this is all about
- [ ] check up on gaps in atomicity


crates/dashchat-node/src/node/app_processing.rs
```diff
@@ -178,26 +194,44 @@ impl Node {
                                 let id = operation.id();
                                 tracing::info!(op = ?id.aliased(), topic = ?topic.aliased(), "application operation processing");
 
+                                // TODO: add accessor to `Acked` to avoid cloning the entire operation.
```
jadehopepunk: It looks like the duplication between  the match branches for `ProcessorEvent::Groups` and `ProcessorEvent::App` has become significant. I'd extract a helper function.

crates/dashchat-node/src/stores/derived_store.rs
```diff
@@ -0,0 +1,262 @@
...
+    "CREATE TABLE IF NOT EXISTS active_inboxes (
+        topic_id BLOB NOT NULL PRIMARY KEY,
+        expires_at_nanos INTEGER NOT NULL
+    )",
+    "CREATE TABLE IF NOT EXISTS group_chats (
+        chat_id BLOB NOT NULL PRIMARY KEY
+    )",
+];
+
+/// The [`DerivedStore`] is a read-only store that is used to make streamlined queries against the [`crate::stores::OpStore`].
```
jadehopepunk: I pondered this name a bit to see if I could poke any holes it it. The word "derived" really clearly implies to me that it's not canonical. 

However, the word name as a whole makes me assume that this is derived from another store (my initial guess was the local store).

However, this store is specifically derived from the operations. That makes it a Projection in event-sourcing speak, so it might be worth considering using that word in the name. Also maybe worth considering removing the word store, since (although it's a database), it's not the canonical store of any data for us.

crates/dashchat-node/src/stores/op_projection.rs
```diff
@@ -0,0 +1,262 @@
+use aliased::Aliasing;
+use p2panda::streams::ProcessedOperation;
+use p2panda_auth::processor::GroupsArgs;
+use sqlx::SqlitePool;
+use std::collections::HashMap;
+
+use crate::{AgentId, DeviceId, Profile, compat::Capabilities};
+use crate::{AnnouncementsPayload, ChatId, ChatPayload, DeviceGroupPayload, Payload, Topic};
+
+const MIGRATIONS: &[&str] = &[
```
jadehopepunk, line 13: This approach to migrations wont always work of course, which is why we typically don't add more migrations to our stores, we currently just increment the database version and make it a breaking change.

However this particular store, being derived, could take a different tack (not necessarily in this PR). You could store the hash of this migration SQL in another table (eg `_schema`) and on boot if it has changed, re-reduce all the operations. That might be slow at some point in the future with large logs, I'm not sure.

It's interesting though that with entirely projected data like this we don't need to have a migration pathway.

crates/dashchat-node/src/stores/local_store.rs
```diff
@@ -69,20 +49,17 @@ impl NodeKeys {
     }
 }
 
+/// The [`LocalStore`] stores only information that is specific to the operation of this device.
+/// It is completely independent and orthogonal to the [`crate::stores::OpStore`] and [`crate::stores::DerivedStore`].
```
jadehopepunk: I don't love the way that claude frequenly references other code in it's comments. This particular sentence referencing the OpStore and DerivedStore should go IMHO. It's incredibly hard to maintain such comments.

crates/dashchat-node/src/stores/op_projection.rs
maackle: I agree with all of this and I'd like it to be a follow up. This code actually changes nothing about the database, just moves the schema and the queries into a different file, and makes the logical split so we can start orienting towards future divergence.

crates/dashchat-node/src/stores/local_store.rs
maackle: Except for a few tests, I actually wrote all of this PR by hand, including this doc ;) I'm surprised you think this should go. I'm trying to identify the role of LocalStore in relation to the other main stores, especially because prior to this PR those concerns were mixed in a single store. I want to make sure we know what goes where in the future. I guess your renaming suggestion would help with that, calling DerivedStore something not about Store, and understanding that "Store" implies source-of-truth rather than projection, but I would still want to see documentation. What would you suggest instead?

crates/dashchat-node/src/node/app_processing.rs
maackle: Actually the only part that made sense to extract was unifying the two acks (p2panda ack and `mark_op_processed`) into a single function, the rest didn't read any better when DRY

crates/dashchat-node/src/stores/derived_store.rs
maackle: Sounds reasonable, and I agree that it is a projection. if you suggest a name, I'll name it that.

maackle:
oof, a not-surprising big conflict with the QR code reduction. I'll have to look at that tomorrow

guillemcordoba (approved):

crates/dashchat-node/src/stores/local_store.rs
maackle: ~~btw rust-analyzer respects doc comment references for refactors, so if we refactor type names, these comments will update too.~~  It's nice for the documentation.

Actually the claim about refactors is false. But it does honor go-to-definition.

maackle:
@jadehopepunk I'm gonna merge this for now and if there's anything you'd like as a followup just let me know

PR #445, opened 2026-07-14
Fix LMS disconnection after mDNS announcement TTL

guillemcordoba:
Both v0.18.9 and v0.19.11 currently have a bug where after 120 seconds of being connected to an LMS, it disconnects. If the app is restarted, the LMS is connected to again.

This PR bumps the mdns-sd crate that includes the fix and also handles network interface changes by itself, removes the network interface watcher at the dash chat level, and also adds a regression test for this case.


e2e-tests/specs/local-mailbox-mdns-ttl.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+		const indicator = agent2.groupChatPage.connectionStatusIndicator;
+		await agent2.waitUntil(
+			async () => (await indicator.status()) === 'local',
+			{ timeout: 90_000, interval: 1_000 },
+		);
+	});
+
+	it('keeps showing the local mailbox icon for 3 minutes, outliving the mDNS TTL', async () => {
+		const indicator = agent2.groupChatPage.connectionStatusIndicator;
+		const deadline = Date.now() + 180_000;
```
maackle, line 87: any way to set the TTL lower for tests to reduce test time?

src-tauri/src/setup.rs
```diff
@@ -59,8 +59,11 @@ pub async fn async_setup(app_handle: AppHandle) -> anyhow::Result<()> {
 
     let _ = crate::APP_HANDLE.set(app_handle.clone());
 
-    // Manage the mDNS service daemon
-    app_handle.manage(mdns_sd::ServiceDaemon::new()?);
+    let mdns = mdns_sd::ServiceDaemon::new()?;
+    if let Err(err) = mdns.set_ip_check_interval(1) {
```
maackle, line 63: I agree with Claude, 5 sec sounds better

maackle (approved):
Would like to see times revisited but otherwise nice fix

src-tauri/src/setup.rs
guillemcordoba: Mmm I actually don't. This interval is really important for the user experience of the connection status indicator. It's a shame it's even an interval, would love to have this be event driven whenever the network interface changes.

Anyway, I think the sacrifice of checking every second to react faster to network changes is very much worth it. Will merge as is for now but let's discuss around it and change it if we decide so.

e2e-tests/specs/local-mailbox-mdns-ttl.spec.ts
guillemcordoba: Unfortunately no... Would need to fork the library.

PR #447, opened 2026-07-14
Simpler edit message frontend logic

guillemcordoba:
I think we should have a requirement that the backend never exposes invalid operations to the frontend. Right now this will not hold as the frontend reads directly from the `OpStore`. But in the migration towards using `p2panda::Node::stream()` that we'll need to do as part of the integration of p2panda-spaces, we should make sure that this requirement holds.


maackle (approved):
Thanks!

PR #449, opened 2026-07-14
Make info exchanged during contact establishment deterministic

maackle:
This doesn't quite work, and I'm not yet sure how to make it right. @jadehopepunk let's talk about it when we sync next


maackle:
This needs more careful consideration as part of our future ACID refactors

PR #450, opened 2026-07-15
e2e tests can run in desktop and android

guillemcordoba:
This PR introduces the affordance of setting the specific platforms in which you want the e2e tests to run, eg:

`AGENT_1=desktop AGENT_2=android just test e2e`

Possible platforms for now are `desktop`, `android` and `android-emulator`. `ios` support will be a follow up PR.


scripts/run-e2e.sh
```diff
@@ -23,5 +25,89 @@ if [[ "$spec_name" != "-" && -n "$spec_name" ]]; then
...
+
+# Build only what the combo needs.
+if $has_desktop; then
+    just test e2e build
+fi
+if $has_android; then
+    # Emulators are x86_64-only; skip the ARM builds unless a physical
+    # device is in the combo.
+    if $has_physical_android; then
+        just test e2e android-build
```
maackle: would be good to still respect a subset of all architectures here for quicker builds if I know e.g. I'm only compiling for aarch64

maackle (approved):
Nice idea!

PR #454, opened 2026-07-16
Ignore compat test and add docs indicating deprecation

maackle:
This one test is our flakiest and longest-running one, and it tests something that I believe we need to rework anyway. So, I propose we ignore it.


crates/dashchat-node/tests/compat_upgrades.rs
```diff
@@ -207,6 +207,7 @@ async fn direct_chat_capability_upgrade() {
 /// - Upgrade cammy; now messages are V1
 /// - A fourth member with zero capability joins; messages revert to V0
 #[tokio::test(flavor = "multi_thread")]
+#[ignore = "TODO: The compatibility system needs to be reworked. Crate-level docs for dashchat-compat."]
```
maackle:
```suggestion
#[ignore = "TODO: The compatibility system needs to be reworked. See crate-level docs for dashchat-compat."]
```

jadehopepunk (approved):
Yep, it's a good idea generally, but right now it's just slowing us down.

PR #457, opened 2026-07-16
Make tombstoning a cross-device operation

maackle:
Mostly moves tombstoning code from LocalStore to OpProjection. Adds the new DeviceGroup payload for tombstoning an operation, so that it's an agent-level rather than device-level decision.

Needed because Delete is also an agent-level decision, and Delete partially relies on tombstoning.


maackle:
Actually, this is probably superceded by Deletes, which are cross-device operations and the only use case for tombstoning.

PR #461, opened 2026-07-21
Block contact UI

maackle:
Quite good blocking UX for us to kick the tires on.

Notes:

- "Reject contact request" is now just replaced with "Block". We have to decide what to do with "Reject" -- maybe that gets wrapped into "Delete Contact"


docs/superpowers/plans/2026-07-20-block-contact-ui.md
```diff
@@ -0,0 +1,745 @@
+# Block-contact UI Implementation Plan
```
guillemcordoba: The plans and specs should not get committed to the codebase.

guillemcordoba (approved):

PR #463, opened 2026-07-21
Add user-reporting backend to mailbox and push servers

maackle:
Introduce a signed `/report` endpoint on both mailbox-server (cloud and local) and push-notifications-server. A report carries a list of device ids, a timestamp, the reporter's pubkey, and the reporter's signature over the ids and timestamp. Servers verify the signature and reject timestamps outside a 1-hour window (const REPORT_TIMESTAMP_TOLERANCE), then store one row per reported device (reporter and timestamp duplicated across rows).

The shared request type, signing-payload construction, timestamp check, signature verification, and row explosion live in a new report-common crate so both servers share one implementation; each server keeps its own storage (redb for the mailbox, sqlx for push).

Clients can send reports: ToyMailboxClient posts to /report, Mailboxes gains report_all to fan out to every connected mailbox, and PushNotificationsClient gains report. Node::report_devices signs a report and sends it to all mailboxes plus the push server (to be wired to the frontend in a followup).


guillemcordoba:
@maackle this can be closed right?

PR #465, opened 2026-07-22
Deep links open direct chats directly

guillemcordoba:


maackle (commented):
It's hard to understand what this is changing, but I don't see any problems

guillemcordoba:
In the previous version of the UI, clicking on a deep link navigated to the add contact view with your own QR, and then immediately took you to the direct chats page, which felt like a glitch. With this change, after clicking a deep link, a loading spinner is shown while processing, and then the appropriate direct chats page is shown.

PR #468, opened 2026-07-24
Add mailbox scrubbing design doc

maackle:
Design for the delete-for-everyone follow-up: publishing payload-free replacements of deleted operations to all mailboxes and dropping their media blobs.

Validation rests on a commitment supplied at store time — the blake3 hash of the operation with its body removed — so the only mutation a stored blip can undergo is the one its publisher pre-authorized.


maackle:
We decided to design scrubbing into the next version of mailbox, not this one for now

PR #470, opened 2026-07-28
Extract DeleteMessageDialog

guillemcordoba:
I propose that in a follow up we introduce the ability to gate Konsta UI animated elements (Sheet, Popover, Dialog) and keep the opening animation. This is why we can't just gate the DeleteMessageDialog with an {#if} in this PR, I think Claude is right about its performance but that right now that's acceptable, as long as we fix it in the future.


maackle:
@guillemcordoba since Claude flagged the e2e tests breaking, I'd lean towards waiting to merge this until it can be pointed at develop and we run e2e tests, and if there is indeed a regression, deciding then whether it makes more sense to fix the tests, or go back to a single dialog and find a way to get the animation back.

PR #474, opened 2026-07-30
Error report goes to sentry

guillemcordoba:
Introduces tauri-plugin-sentry-reporting, and wires the send error report functionality to sentry. This has been tested end-to-end and errors get to our sentry dash chat project.

Usage of tauri-plugin-sentry has been considered but decided against because we need control over the redaction of the logs and the moment in which they are sent: we want to be able to only send logs when the user chooses to do so.

Sentry is activated using the SENTRY_DSN environment variable.

This is the first vertical slice of the improving user feedback story.


crates/tauri-plugin-sentry-reporting/src/commands.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    // Reading a megabyte of log and running the patterns over it — and over the
+    // event itself — is blocking and CPU-bound, so it stays off the async worker.
+    tauri::async_runtime::spawn_blocking(move || send(&state, event))
+        .await
+        .map_err(|err| format!("Report task failed: {err}"))
+}
+
+/// Enriches, redacts and transmits the report. The only path to the network.
+fn send(state: &SentryState, event: Event<'static>) {
```
jadehopepunk, line 58: There's no error return here?

crates/tauri-plugin-sentry-reporting/src/redaction.rs
```diff
@@ -0,0 +1,207 @@
...
+    patterns: &[Regex],
+    event: Event<'static>,
+) -> anyhow::Result<Event<'static>> {
+    let mut value = serde_json::to_value(event)?;
+    redact_json_leaves(patterns, &mut value);
+    Ok(serde_json::from_value(value)?)
+}
+
+/// Tails to `max_bytes`, dropping the leading partial line.
+pub(crate) fn read_concat_tail(paths: &[PathBuf], max_bytes: usize) -> std::io::Result<String> {
```
jadehopepunk: The function name could be tweaked to better convey what's in the comment. I think the "dropping the leading partial line" is a key part. Maybe `fn tail_whole_lines` which assumes knowledge of using "tail" as a verb (ie the tail command) and also assumes it's read it context with the max_bytes param.

crates/tauri-plugin-sentry-reporting/src/redaction.rs
```diff
@@ -0,0 +1,207 @@
...
+        if let Some(pos) = buf.find('\n') {
+            buf.drain(..=pos);
+        }
+    }
+    Ok(buf)
+}
+
+/// Oldest first. KeepOne rotation leaves a date-stamped sibling next to the live
+/// file, and the report wants both.
+pub(crate) fn list_log_files_oldest_first(dir: &Path) -> std::io::Result<Vec<PathBuf>> {
```
jadehopepunk, line 72: "KeepOne rotation leaves a date-stamped sibling next to the live file, and the report wants both" is not a comment pertaining to this function. It's describing a situation in which this function is used. I find that Claude does this a lot, and I really doubt it's ability to keep those comments true as functions become used in other ways or their context changes.

Certainly it can be nice to have a "why" in a comment for a function or module, but the why needs to be abstracted for all potential users, eg "In case there are multiple data-stamped log files in the same dir, perhaps as a result of log rotation tools".

jadehopepunk (approved):
Looks great, I really like the introduction of the crate, and the place where you separated between the general log handling code and the app specific redaction regexes. 

Some small nits commented on regarding naming and comments, but I don't feel strongly about them.

crates/tauri-plugin-sentry-reporting/src/redaction.rs
guillemcordoba: Agreed, thank you. In this case these functions were only moved from the redact_log.rs file to this file, so didn't put attention into cleaning them up. Will clean up the comments.

crates/tauri-plugin-sentry-reporting/src/commands.rs
guillemcordoba: The built-in transport doesn't actually tell us if there's been an error or not. A follow-up PR will replace the built-in transport with a custom one with which we can make that differentiation.

PR #476, opened 2026-07-31
Proof-of-concept android "background" service.

jadehopepunk:
This PR introduces a way of doing android background services. The service doesn't "do" anything, it just increments a number and logs it. It is started with a link in the help section, which only displays on android, and only in dev builds.


ui/src/routes/settings/help/+page.svelte
```diff
@@ -43,6 +45,18 @@
 					title={m.contactUs()}
 					data-testid="help-contact-us"
 				/>
+				{#if isAndroid && import.meta.env.DEV}
+					<ListItem
+						link
+						chevron={false}
+						title="Start offline mode"
```
guillemcordoba: Title should be internationalized.

guillemcordoba (approved):

PR #477, opened 2026-07-31
Record last app logs as Sentry logs

guillemcordoba:
Collects app logs as Sentry logs. 

Only the latest app logs get submitted as Sentry logs for indexing and easy searching between issues, the full log file gets added to the report as an attachment.

Also refactors the way we set up Sentry to reuse as much of their built-in integrations as possible, instead of having custom sentry code on our side.


crates/tauri-plugin-sentry-reporting/src/client.rs
```diff
@@ -1,49 +1,61 @@
...
-    }
-    event
-        .contexts
-        .entry("rust".into())
-        .or_insert_with(rust_context);
-    event
-        .contexts
-        .entry("device".into())
-        .or_insert_with(device_context);
+/// Sentry's own pipeline, minus everything that would transmit unasked.
```
jadehopepunk: This comment is pretty confusing. The function looks good though

crates/tauri-plugin-sentry-reporting/src/client.rs
```diff
@@ -52,29 +64,41 @@ mod tests {
...
+
+    fn options_keeping(pending: Arc<PendingLogs>) -> ClientOptions {
+        let (gate, _) = recording_gate();
+        options(&config(PathBuf::new()), pending, Arc::new(gate))
     }
 
     #[test]
-    fn the_sdk_can_never_transmit_on_its_own() {
-        let before_send = options(&config()).before_send.unwrap();
+    fn a_captured_log_is_kept_redacted_rather_than_queued_for_sending() {
```
jadehopepunk, line 80: I love the line breaks in this test, splitting it into setup, function under test, and observed results. :)

crates/tauri-plugin-sentry-reporting/src/envelope.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let mut envelope: Envelope = event.into();
+    if !logs.is_empty() {
+        envelope.add_item(logs::envelope_item(logs, trace_id));
+    }
+    Some(envelope)
+}
+
+/// The only path to the network. The logs and the log file overlap on purpose:
+/// the logs are searchable but recent, the file reaches further back.
```
jadehopepunk: The comment here is confusing. There is something to the "the only path to the network", although it's hard for a comment to ensure that.

Perhaps "The only way  to send a report to sentry" would convey that fact plus the purpose of the function. I'm not sure what the rest of the comment is for.

crates/tauri-plugin-sentry-reporting/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+        event.exception = vec![Exception {
+            ty: error.name,
+            value: Some(error.message),
+            ..Default::default()
+        }]
+        .into();
+    }
+
+    let logs = state.pending.snapshot();
+    if let Some(envelope) = envelope::build_envelope(&state, event, logs) {
```
jadehopepunk, line 40: What happens if build_envelope returns None?

crates/tauri-plugin-sentry-reporting/src/logs.rs
```diff
@@ -0,0 +1,59 @@
...
+use sentry::integrations::log::log_from_record;
+use sentry::protocol::{EnvelopeItem, ItemContainer, Log, TraceId};
+use tauri::{AppHandle, Manager, Runtime};
+use tauri_plugin_log::fern::{Dispatch, Output};
+use tauri_plugin_log::{Target, TargetKind};
+
+use crate::state::SentryState;
+
+/// The attachment holds the whole history, so this only has to span the window
+/// worth indexing
```
jadehopepunk: This comment is pretty confusing given that it's on a MAX_ENTRIES const. What attachment. How would we index a MAX_ENTRIES?

jadehopepunk (approved):
Looks good. Commented with some minor suggestions, mostly about comments.

PR #489, opened 2026-08-04
Publish attempts immediate sync

guillemcordoba:
Without this, `trigger_sync()` in media waited for the next poll cycle before actually syncing with the mailboxes.


maackle (changes requested):
I think this is good for now but immediate sync, especially to degraded connections, ideally would be debounced. I think at minimum, consider making it `due_now_if_active` instead of `if_not_stopped`, so it skips degraded too. A mailbox has to be having some consistent errors to be in degraded status so it is likely that an immediate sync would fail anyway, and I imagine the whole point of this is to create a quicker experience for healthy mailboxes (a degraded mailbox is already going to be a slow experience due to it being degraded, that can't be fixed by the client).

PR #491, opened 2026-08-05
Convert lots of usage of Topic<Untyped> to just TopicId

maackle:


guillemcordoba (approved):
I like the simplicity this brings.

PR #492, opened 2026-08-05
Refactor/access to node

jadehopepunk:


guillemcordoba (approved):

guillemcordoba:
@jadehopepunk tested in iOS backgrounding and foregrounding the app, and also receiving a push notification while the app is in running and not: working as intended, no issues encountered.

PR #494, opened 2026-08-05
Camera button in select avatar opens camera

guillemcordoba:
The camera button in the select avatar page was wrongly showing the photo gallery, exactly like the "Gallery" button.

This PR opens the camera and sets the captured image as the avatar.

### DIFF
--- a/ui/src/lib/utils/image.ts
+++ b/ui/src/lib/utils/image.ts
@@ -28,3 +28,22 @@ export function resizeAndExport(
 	// return the .toDataURL of the temp canvas
 	return canvas.toDataURL();
 }
+
+/**
+ * Decode an image file and export it as an avatar-sized data URL. Rejects if
+ * the file cannot be read or is not an image the webview can decode.
+ */
+export function fileToAvatar(file: File): Promise<string> {
+	return new Promise((resolve, reject) => {
+		const reader = new FileReader();
+		reader.onerror = () => reject(new Error(`failed to read ${file.name}`));
+		reader.onload = e => {
+			const img = new Image();
+			img.crossOrigin = 'anonymous';
+			img.onerror = () => reject(new Error(`failed to decode ${file.name}`));
+			img.onload = () => resolve(resizeAndExport(img));
+			img.src = e.target?.result as string;
+		};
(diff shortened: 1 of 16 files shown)


ui/src/lib/utils/image.ts
```diff
@@ -28,3 +28,22 @@ export function resizeAndExport(
 	// return the .toDataURL of the temp canvas
 	return canvas.toDataURL();
 }
+
+/**
+ * Decode an image file and export it as an avatar-sized data URL. Rejects if
```
maackle, line 33: why do we use data URLs and not byte arrays?

ui/tests/setup-utils.ts
```diff
@@ -199,6 +199,54 @@ function dropFiles(specs: TestFileSpec[]) {
...
+const nativeInputClick = HTMLInputElement.prototype.click;
+let filePickerRequests: FilePickerRequest[] | undefined;
+
+/**
+ * Record the file inputs the app opens instead of letting them reach the OS,
+ * answering each with `files` — or, when there are none, backing out of it as
+ * if the user had dismissed the dialog.
+ */
+function interceptFilePickers(files: TestFileSpec[] = []): void {
+	filePickerRequests = [];
```
maackle, line 219: seems that this array is never used? was it for logging?

maackle (approved):
Looks pretty good and I have 2 questions

ui/tests/setup-utils.ts
guillemcordoba: It is used here: https://github.com/dash-chat/dash-chat/pull/494/changes#diff-af97d5209bb85a5e27e4c50299589b4ae466d0c1df8201e5c2e8645cc73c2cedR245

ui/src/lib/utils/image.ts
guillemcordoba: It's what `` takes, and it also allows the flexibility to define our own avatar formats like the [text-avatar-url](https://github.com/dash-chat/dash-chat/blob/develop/ui/src/lib/components/profiles/text-avatar-data-url.ts) while still just having a simple `avatar: Option<String>` field.

PR #495, opened 2026-08-05
TombstoneStore

maackle:


mprocs.yaml
```diff
@@ -1,8 +1,8 @@
 procs:
   agent1:
-    shell: DATA_DIR="$PWD/.dbs/dev/agent-1" just dev tauri
+    shell: DATA_DIR="$PWD/.dbs/dev/agent-1" just dev tauri 2>&1 | tee agent1.log
```
guillemcordoba, line 3: I think this doesn't belong here.

guillemcordoba (approved):
Looks really good! Really happy to have the TombstoneStore as a nice pattern in the code now. Just a small comment.

mprocs.yaml
maackle: @guillemcordoba yeah I'll take it out for now but I think we need a better way to access mprocs logs. They're not even greppable AFACIT

PR #496, opened 2026-08-05
Fix background flash

guillemcordoba:
Right now, if the system theme is dark, when Dash Chat is opened, the webview background color flashes white before the actual app is rendered.

This PR fixes this issue by moving all theme handling functionality to [tauri-plugin-system-theme](https://github.com/dash-chat/tauri-plugin-system-theme) which applies the system theme automatically on startup. This PR refactors the app to use its new capabilities.


ui/src/lib/utils/theme.ts
```diff
@@ -1,38 +1,23 @@
...
 export async function setLightSystemBars() {
-	if (isMobile && isTauriEnv()) {
-		await setSystemBarsStyle('light', 'light');
-	}
+	if (!isMobile || !isTauriEnv()) return;
+	await overrideSystemBarsColorScheme('light');
 }
 
-/** Restore the system-bar style to match the currently-applied theme. */
+/** Give the system bars back to the app's colour scheme. */
```
maackle: Confusing, this instead?

```suggestion
/** Make the system bars match the app's colour scheme. */
```

maackle (approved):

PR #501, opened 2026-08-07
Onboarding welcome screen

guillemcordoba:
Adds a welcome and a EULA screen.


maackle (changes requested):
It does seem we'll need to localize the entire EULA, presumably with https://paraglidejs.com/i18n-routing ?

PR #502, opened 2026-08-07
Reporting

maackle:


guillemcordoba (commented):
I think this assumes reporting can only happen when the user is connected to some mailbox. Only then the `ReportContact` operation is created.

Signal allows you to "mark user as reported" even when offline. I think we should do the same if possible, and invert the current flow: when the user clicks report, the report operation is published, and there is some retry mechanism that keeps retrying to send the reports to the mailboxes that haven't gotten them, and probably would publish an operation to the device group announcing its success. 

I would be fine if that meant that for now we're only reporting to the cloud mailbox server.

guillemcordoba (approved):

PR #505, opened 2026-08-10
useReactivePromise refactor

maackle:
actually based on `fix/delete-for-me`, but wanting to run on CI.

Claude seemed to think this would improve our e2e testing, since useReactivePromise leads to components repeatedly being unmounted and replaced. I haven't yet verified that claim.


guillemcordoba:
@maackle closing this as already fixed in the delete for me branch.

PR #508, opened 2026-08-10
Replies

maackle:


e2e-tests/helpers/components/contact-actions-menu.ts
```diff
@@ -1,9 +1,11 @@
 import { TestHelper } from '../pages/test-helper';
 import { tid } from '../selectors';
 
-/** The contact overflow popover and the block dialog it opens. */
+/** The contact overflow popover and the block/report dialogs it opens. */
```
maackle: note to self: resume the review here

also, double check validation and rendering of replies to edit chains

crates/dashchat-node/src/chat/message.rs
```diff
@@ -1,29 +1,18 @@
...
-}
-
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct ChatMessageContentV1 {
     pub message: String,
     pub media: Option<MediaBundle>,
+    /// Hash of the operation this message replies to: a `Message` or
+    /// `EditMessage` in the same topic (any author's log). When the target has
+    /// been edited, an honest node replies to the latest edit it knows of.
+    /// Absent on the wire for non-replies, so old clients keep decoding V1
```
guillemcordoba, line 11: I think the skip_serializing is not relevant for our current state and should be removed.

packages/stores/src/chats/messages-store.ts
```diff
@@ -91,6 +96,13 @@ export class MessagesStore {
 		for (const { hash, reason } of tombstones) {
 			if (reason === 'DeletedForMe') delete messages[hash];
 		}
+
+		const ops = collectOps(logs);
+		const deletedForMe = deletedForMeHashes(tombstones);
+		for (const [hash, message] of Object.entries(messages)) {
+			messages[hash] = applyReply(message, messages, ops, deletedForMe);
```
guillemcordoba: This is doing too much work: instead of going through all messages again, we should add replies to the return type of collectMessagesByActionType, and do the processing only of those here.

guillemcordoba: This is doing too much work: instead of going through all messages again, we should add replies to the return type of collectMessagesByActionType, and do the processing only of those here.

packages/stores/src/chats/replies.ts
```diff
@@ -0,0 +1,219 @@
...
+interface OpInfo {
+	author: DeviceId;
+	timestamp: number;
+	body: Payload | undefined;
+}
+
+/** Every operation in the chat's logs by hash, the lookup reply resolution
+ * walks: quotes and scroll targets point at ops that may not render as
+ * messages themselves (edits, tombstoned bodies). */
+export function collectOps(
```
guillemcordoba: Why does this exist? We already have collectMessagesByActionTypes, we don't want to need to walk all the messages twice.

ui/src/lib/components/contacts/ContactActionsMenu.svelte
```diff
@@ -38,6 +40,13 @@
 			onClick={() => (blockDialogOpen = true)}
 			data-testid="contact-block"
 		/>
+		<ListAction
+			title={m.report()}
```
guillemcordoba: I'm confused, why is this list action item here?

ui/src/lib/components/messages/composer/ReplyBanner.svelte
```diff
@@ -0,0 +1,50 @@
...
+	<wa-icon src={wrapPathInSvg(mdiReply)} style="font-size: 0.9rem"></wa-icon>
+	<span class="column min-w-0 flex-1">
+		<span class="truncate font-semibold">
+			{m.replyingTo({ name: authorName })}
+		</span>
+		<span class="quiet truncate" data-testid="composer-reply-preview">
+			{preview}
+		</span>
+	</span>
+	<button
```
guillemcordoba: Use IconButton instead of raw button+wa-icon

packages/stores/src/chats/messages-store.ts
```diff
@@ -38,6 +39,10 @@ export interface Message {
 	timestamp: number;
 	author: DeviceId;
 	seqNum: number;
+	/** Raw hash of the operation this message replies to, straight from the log. */
```
guillemcordoba: I think we should unify these two fields.

packages/stores/src/chats/replies.ts
```diff
@@ -0,0 +1,219 @@
...
+	return reply;
+}
+
+/** Resolve the message's reply annotation and return the resulting message.
+ * Must run after edits, deletes and delete-for-me tombstones have been applied
+ * to `messages`: a quote pointing at a message that ended up deleted falls
+ * back to a tombstone, and its scroll target is only kept when the
+ * corresponding placeholder actually renders. */
+export function applyReply(
+	message: Message,
```
guillemcordoba: I think this should go as part of the refactor of collectMessagesByActionType

guillemcordoba (changes requested):
I think this is in the right direction, I think the frontend store logic side should be improved.

ui/src/lib/components/contacts/ContactActionsMenu.svelte
maackle: I think it was missing before so it's getting snuck in here.

ui/src/routes/group-chat/[chatId]/+page.svelte
```diff
@@ -144,6 +152,34 @@
 
 		return { hash: capturedUnreadHash, count };
 	}
+
+	function navigateToMessage(hash: Hash) {
+		scrollToMessage(messagesEl, hash);
+	}
+
+	function deviceDisplayName(
```
Comment (guillemcordoba) [inline ui/src/routes/group-chat/[chatId]/+page.svelte:160]: I think this should not be duplicated across group chat and direct chat pages, I think it should be moved to the MessageComposer.

ui/src/routes/group-chat/[chatId]/+page.svelte
```diff
@@ -144,6 +152,34 @@
...
+		if (deviceId === myDeviceId) return m.you();
+		const member = Object.values(members).find(m =>
+			m.deviceIds.includes(deviceId),
+		);
+		return member?.profile ? fullName(member.profile) : m.unknownSender();
+	}
+
+	/** Display name of the author quoted by `message`'s reply, if that author is
+	 * known — a quote of a message this peer never received has none. */
+	function quotedAuthorName(
```
Comment (guillemcordoba) [inline ui/src/routes/group-chat/[chatId]/+page.svelte:174]: Same, I think this should be moved down to its correspondant elements.

ui/src/lib/components/messages/message-helpers.ts
```diff
@@ -67,6 +67,71 @@ export function highlightMatch(text: string, query: string): string {
 	);
 }
 
+/** Run `callback` once `el` has stopped moving under a smooth scroll, or right
+ * away if it never starts moving (target already in view, or the container is
+ * already at its scroll limit). Tracks the element's viewport position rather
+ * than a container's scrollTop so it doesn't need to know which ancestor
+ * scrolls. */
+function afterScrollSettles(el: Element, callback: () => void): void {
```
guillemcordoba: This is an animation callback function that runs at every step of the animation, which is potentially harming performance on low resource devices. Is there a way to do it without running logic at each step?

guillemcordoba (changes requested):
Looking really good! I see 3 relevant improvements, after them it should be good to go.

guillemcordoba (approved):

PR #511, opened 2026-08-11
Remove compat versioning

maackle:
More in the vein of #455


jadehopepunk (approved):

maackle:
Broken here but fixed in #508, will be merged in with that.

PR #512, opened 2026-08-11
Implement clickable links

guillemcordoba:
Render URLs in text messages as clickable links, that open in the default browser.


jadehopepunk (approved):
This looks good. The changes made to make it testable look fine. The use of linkifyjs is sound.

This is totally suitable for this feature set, but one thing to consider might be supporting markdown. Matrix clients typically do it, although they don't say it anywhere. Normal users don't notice (they just get link highlighting and stuff) but power users can do nicer things.

PR #513, opened 2026-08-11
Refactor media types

guillemcordoba:
This is done to be able to accomodate https://github.com/dash-chat/dash-chat/pull/516.

The new voice note attachment type does not fit cleanly with the generic Media struct.


maackle (commented):
Looks fine, but per https://github.com/dash-chat/dash-chat/pull/516/changes#r3762025620 this may not be necessary depending on the waveform/duration decision

PR #516, opened 2026-08-11
Add voice note type to backend and frontend

guillemcordoba:


crates/dashchat-node/src/chat/message.rs
```diff
@@ -43,8 +43,17 @@ pub struct OutgoingFile {
...
 
-/// Media attached to a chat message. A message has either a set of photos
-/// or a single file — not both — matching Signal's UX.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct OutgoingVoiceNote {
+    pub data: Vec<u8>,
+    pub mime_type: String,
+    pub duration_ms: u32,
+    // Amplitude bars (`0..=255`) for the UI
+    pub waveform: Vec<u8>,
```
maackle, line 52: Seems wasteful! Let the receiver render their own waveform

crates/dashchat-node/src/chat/message.rs
```diff
@@ -86,12 +97,22 @@ pub enum MediaMetadata {
         #[serde(with = "hash_bytes")]
         hash: iroh_blobs::Hash,
     },
+    VoiceNote {
+        mime_type: String,
+        size: u64,
+        duration_ms: u32,
+        waveform: Vec<u8>,
```
maackle, line 104: also not needed

maackle (changes requested):

crates/dashchat-node/src/chat/message.rs
maackle: duration also seems unnecessary and can easily be calculated. is there a reason we need that in the metadata?

if unnecessary, then #513 is also unnecessary

guillemcordoba:
@maackle these properties are necessary to render the proper UI for voice notes even if we haven't downloaded the voice note blob. I think this is desirable:

1) In the connected to the internet case, you want to present to the user the message as soon as possible, and showing only the text messages and then when the audio has finished downloading add to those the audio message feels like a really bad UX.
2) In the not-connected-to-the-internet case, even if you don't have the blob itself, I think you do want to see that the audio message exist, its duration and waveform.

guillemcordoba:
And actually, I think we need to extend the photo attachment type to include height/width. Otherwise the frontend does not know them before downloading the blob, leading to a bad UX: I see the photo as a generic square, and then when the image is downloaded, it changes size and moves the layout around.

maackle (approved):
I wasn't really thinking this through all the way. Now that I have, I have new observations:

Signal doesn't seem to make use of this kind of metadata. When I receive a long voice note, I don't see duration or waveform, and the clip doesn't even download until I press play. I do think we can do better though, as you say, but if we're going for strict parity then it actually isn't necessary.

Your point about extending image metadata with dimensions is actually a more compelling argument for media metadata variants, since we are deviating from Signal in the regard of fetching images individually rather than as a batch.

I'm fine with this now, I think we should improve upon Signal where it's clear that we can. Maybe have one more quick think about any possible security implications of including that metadata vs not. I can't think of anything significant enough to matter.

PR #519, opened 2026-08-12
Stop the toast from swallowing clicks in the bottom corners

maackle:
Fixes the remaining e2e failure on #502 ([run 31544297578](https://github.com/dash-chat/dash-chat/actions/runs/31544297578/job/93953367868)), where `qr-upload` fails with `element click intercepted` on `home-new-message-btn`.

## Why the click is intercepted

Konsta's toast root is not the visible pill — it is a full-width bar pinned across the bottom of the screen with the pill centred inside it:

```
k-toast z-50 … w-auto start-0 bottom-0 fixed left-safe-4 right-safe-4 bottom-safe-4 flex justify-center
```

At `z-50` it wins the hit test against the new-message FAB (`z-20` inside a `z-10` container) in the bottom corner. `ToastManager` disabled pointer events on that root for every variant *except* `unexpected` — and `unexpected` is also the one variant that never auto-dismisses (`ToastManager.svelte:27` only arms the TTL for the others), because the user needs to be able to reach "Send error report". So a single unexpected-error toast blanks out the bottom corners of the app for the rest of the session.

In the failing spec that toast is raised during the *successful* QR-upload contact add in the preceding test — an unhandled promise rejection, logged as `[unhandledrejection]` immediately after `add_contact` publishes — and is still up when the next test clicks the FAB. The uploaded failure screenshot shows the chat list with the toast parked at the bottom and the FAB unreachable.

## The change

Keep the toast root click-through for every variant, and re-enable pointer events on the error controls, which are the only part that needs them. The message text and the pill's padding no longer take clicks — matching what non-`unexpected` toasts already did — while "Send error report" and the close button keep working.

## Scope

This fixes the interception, not the spurious rejection behind it. The app raising "An unexpected error occurred." after a *successful* QR-code contact add is a separate, user-facing bug still worth chasing: the reason is an `Error` with an empty message one frame into minified `chunks/…js`, and `addContactFromCode` (`ui/src/lib/deep-links/add-contact.ts:42`) try/catches its whole body, so it is a floating promise somewhere off that path. Resolving it needs a local repro with source maps.

## Verification

The component compiles clean under the Svelte 5 compiler with no warnings, and is Prettier-clean. I could not run the e2e suite or the app: `pnpm install` cannot complete in this environment (the private `tauri-plugin-system-theme` / `tauri-plugin-ios-photos` / `tauri-plugin-virtual-keyboard` git dependencies are not reachable from this session), so CI on this PR is the real check.

## Also seen in that run, not addressed here

`send-messages` failed its last four tests on the first attempt (`Message "Hello from Bob!" not found`, `Composer did not clear after sending …` twice, then `direct-chat-search-input` never appearing) and passed clean on retry. It did not turn the run red, but four consecutive failures followed by a clean rerun looks more like a real sync/timing problem than incidental flake.

---
_Generated by [Claude Code](https://claude.ai/code/session_012CbuaVqPMZ63yLVtJyEkm8)_

### DIFF
--- a/e2e-tests/specs/report-contact.spec.ts
+++ b/e2e-tests/specs/report-contact.spec.ts
@@ -26,10 +26,15 @@ describe('report contact', () => {
 		// One-way add, so Alice lands on a chat showing Bob's contact request.
 		await navigateToAddContact(agent1);
 		const aliceLink = await agent1.addContactPage.getAddContactLink();
+		await agent1.addContactPage.back.click();
+		await agent1.newMessagePage.back.click();
+		await agent1.homePage.ready();
+
 		await navigateToAddContact(agent2);
 		await agent2.addContactPage.enterAddContactLink(aliceLink);
 		await agent2.directChatPage.ready();
 
+		await agent1.homePage.chatList.waitForExist();
 		await agent1.homePage.openChat('Bob Test');
 		await agent1.directChatPage.acceptButton.waitForExist();
 	});
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -29,6 +29,7 @@
 	"edited": "Edited",
 	"editingMessage": "Edit message",
 	"deleteMessageTitle": "Delete message?",
+	"deleteMessageDescription": "Who would you like to delete this message for?",
 	"deleteForEveryone": "Delete for everyone",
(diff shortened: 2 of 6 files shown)


maackle:
This runs in CI successfully and looks good in testing so I'm merging.

PR #522, opened 2026-08-12
Delayed search flash

maackle:
In Replies, when scrolling up a long way to an older message, the flash can be over by the time the scroll completes. This delays the flash until the scroll has settled. Also adjusts the brightness and duration of the flash to match Signal better.

### DIFF
--- a/ui/src/app.css
+++ b/ui/src/app.css
@@ -297,23 +297,30 @@ li.plain span {
 }
 
 .search-flash {
-	animation: search-flash-anim 1.5s ease-out;
+	--flash-brightness: 1.5;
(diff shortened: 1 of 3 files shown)


guillemcordoba (commented):
I like what the goal of this PR.

I don't like the way it achieves it, specifically the manual frame-by-frame animation. This is the kind of code that is intensive and not performance friendly. Before conceding that we need to accept this as the best solution, I would like other solutions that are not frame-by-frame explored. Can you do that @maackle ?

guillemcordoba:
@maackle can we close this?

maackle:
already mostly handled in the other replies work

PR #527, opened 2026-08-13
Generalize ActionDialog to accept multiple actions

guillemcordoba:
New dialogs like delete message dialog or the report contact dialog need 2 actions other than cancel. This PR generalizes ActionDialog to accept an array of actions.


jadehopepunk (approved):
Love it 😁

PR #531, opened 2026-08-16
Simplify message rendering

maackle:


maackle:
@guillemcordoba I went back and rewrote the edit/delete/reply rendering to fit in a single function. It's a big function but I think much clearer and more straightforward.

guillemcordoba (approved):

PR #536, opened 2026-08-20
LMS docker image

guillemcordoba:
Introduces a docker image to run local message servers. 

This represents just the initial work to make them work. In follow-up PRs I'll refactor the way that the mailbox server and local mailbox server crates are structured so that we can deduplicate the logic.


maackle (approved):
Looks fine, want to do that TODO now though?

PR #537, opened 2026-08-23
Change background mode to be server-side initiated

jadehopepunk:
- Renamed Offline Mode to Background Mode
- Changes to using the general setting store rather than a specific store
- Use new version of tauri-lifecycle-plugin
- Initiate the android "foreground task" when switching out of the app according to background lifecycle events


src-tauri/src/background/lifecycle.rs
```diff
@@ -0,0 +1,28 @@
...
+use tauri::{AppHandle, Wry};
+use tauri_plugin_background_service::ServiceManagerHandle;
+
+pub(crate) async fn on_pause(app: AppHandle<Wry>) {
+    if !crate::settings::load_settings(&app).background_mode_enabled {
+        return;
+    }
+    let manager = app.state::<ServiceManagerHandle<Wry>>();
+    let config = tauri_plugin_background_service::StartConfig {
+        service_label:
```
guillemcordoba: This label should be i18n'd, we do have that ability in the backend.

ui/src/routes/settings/offline/+page.svelte
```diff
@@ -54,26 +58,52 @@
...
-			</List>
-			<BlockFooter class="px-4">{m.localMessageServerDescription()}</BlockFooter
-			>
-		</div>
+			</div>
+		{/if}
+
+		{#if isAndroid}
+			<div class="column center-in-desktop">
+				<BlockTitle>Background Mode</BlockTitle>
```
guillemcordoba: Background mode should also be i18n'd.

guillemcordoba (approved):
Just two i18n change requests, otherwise looks great!

PR #539, opened 2026-08-23
Frontend/early direct chat

guillemcordoba:


crates/dashchat-node/src/node.rs
```diff
@@ -1639,10 +1645,17 @@ impl Node {
...
+        &self,
+        agent_id: AgentId,
+        direct_chat_topic_id: ChatId,
+    ) -> Result<(), Error> {
         self.publish(
             self.device_group_topic(),
-            Payload::DeviceGroup(DeviceGroupPayload::AddContact { agent_id }),
+            Payload::DeviceGroup(DeviceGroupPayload::AddContact {
+                agent_id,
+                direct_chat_topic_id,
```
maackle, line 1659: good, this makes it easier to change to a random topic id if we want in the future

crates/dashchat-node/Cargo.toml
```diff
@@ -63,11 +63,7 @@ tempfile = "3"
...
 [features]
-default = ["instrument", "fake-agent-id"]
+default = ["instrument"]
 instrument = []
 testing = ["tempfile", "nanoid", "p2panda-store/test_utils"]
 p2p = []
-
-# Represents our current false equivalence between DeviceId and AgentId.
-# This must be removed and cleaned up before device groups can work.
-fake-agent-id = []
```
maackle, line 73: you didn't like this pattern? I *want* there to be ugly things to remind us of the hacks we need to clean up.

packages/stores/src/contacts/contacts-store.ts
```diff
@@ -5,27 +5,40 @@ import { LogsStore } from '../p2panda/logs-store';
...
+export interface Contact {
+	agentId: AgentId;
+	chatId: ChatId;
+	addedTimestamp: number;
+}
+
+export interface ContactWithProfile {
+	contact: Contact;
+	profile: Profile;
+}
```
maackle, line 40: Was it necessary to distinguish these? I wondered if the truncated name could be the basis of the "profile" from the first moment.

packages/stores/src/mock/client.ts
```diff
@@ -19,6 +19,11 @@ export function hash<T>(obj: T): Hash {
 	return blake2bHex(JSON.stringify(obj));
 }
 
+/** Mirrors the backend's direct-chat topic derivation from the two device ids. */
+export function chatIdForDevices(deviceA: string, deviceB: string): string {
+	return hash([deviceA, deviceB].sort().join(':'));
+}
```
maackle, line 24: I get that this is a mock that doesn't interact with the backend but it might be better to just not hash this concatenated string, so it's clearer that it's a mock value. This seems a bit uncanny valley atm.

maackle (approved):
Looks good, with a few comments

crates/dashchat-node/Cargo.toml
guillemcordoba: Okey! just didn't want more complexity than absolutely necessary, but it's fine I'll readd it.

packages/stores/src/contacts/contacts-store.ts
guillemcordoba: Yeah good point, I don't think the distinction has enough benefits, I'll add profile to the Contact interface.

guillemcordoba: After more thorough thought, there is a phase in which we don't have the profile for a contact, just after scanning a QR Code. So I'll keep this distinction for now, we can revisit in the future.

PR #541, opened 2026-08-24
Add "secret" developer settings

maackle:
Tap 7 times on the Version line in Settings to unlock a new (empty) Developer settings item


ui/src/lib/stores/developer-mode.svelte.ts
```diff
@@ -0,0 +1,22 @@
+const STORAGE_KEY = 'developer-mode-unlocked';
```
guillemcordoba, line 1: I think this should actually go in the settings, so that when we do "delete account" developer mode also gets reset.

guillemcordoba (approved):
Looks great, I'd rather the developer mode be persisted in the settings store.

PR #542, opened 2026-08-24
Store sentry crashes, reports, feedback in a file-backed queue for sending when network connection returns

maackle:
All sentry reports can now be initiated offline. All outgoing Sentry stuff is stored in a filesystem-based queue for resilient delivery. When offline, or even while online with a retryable send, the queue will attempt to be drained until success, with backoff. The queue structure is implemented entirely by sorted filenames. Standard measures are taken to avoid inconsistency and disk corruption, like saving to a temp file first and then renaming files.

[docs/sentry-reporting.md](https://github.com/dash-chat/dash-chat/blob/2aede45e82696b7c6b88a194ce9aad23be220807/docs/sentry-reporting.md) has a detailed overview of the system

## Refactors:

- moves network changed/settled detection from dashchat-node into dashchat-utils
- simplifies `UserInitiatedTransport` to not use an internal nested Transport. It just is a Transport.


crates/dashchat-utils/Cargo.toml
```diff
@@ -6,8 +6,9 @@ edition = "2021"
 [dependencies]
 anyhow = { workspace = true }
 async-trait = { workspace = true }
-iroh = { workspace = true }
-iroh-blobs = { workspace = true }
+iroh = { workspace = true, optional = true }
+iroh-blobs = { workspace = true, optional = true }
+if-watch = { version = "3", features = ["tokio"] }
```
maackle, line 11: maybe it's undesirable to include more deps and code in anything else that uses dashchat-utils?

docs/sentry-reporting.md
```diff
@@ -0,0 +1,305 @@
+# Sentry Error Reporting
```
guillemcordoba, line 1: I think we should remove this file: it reads mostly as a plan for this PR, and we'd like to have the plugin's documentation live in its README.

crates/dashchat-utils/src/network_settled.rs
```diff
@@ -0,0 +1,143 @@
...
+
+const SETTLE: Duration = Duration::from_millis(1500);
+const CAPACITY: usize = 4;
+
+/// Subscribes to debounced network-settled events.
+///
+/// The first call spawns the single interface watcher; later calls hand out
+/// further receivers of the same signal.
+pub fn network_settled() -> broadcast::Receiver<()> {
+    static SENDER: OnceLock<broadcast::Sender<()>> = OnceLock::new();
```
guillemcordoba, line 26: Nice trick!

crates/tauri-plugin-sentry-reporting/src/state.rs
```diff
@@ -2,35 +2,182 @@ use std::path::PathBuf;
...
+    }
+}
+
+/// What to tell the user, and Sentry, about a report that went nowhere.
+fn explain(why: DropReason) -> String {
+    match why {
+        DropReason::Refused {
+            status: Some(status),
+        } => {
+            format!("Sentry refused the report: {status}")
```
guillemcordoba, line 43: These would need to be i18n if they eventually get to the user.

guillemcordoba (approved):
Looking really really good, small comments.

crates/tauri-plugin-sentry-reporting/src/state.rs
maackle: Looks like these never do reach the user, they only go in Sentry. So I think we're good.

PR #544, opened 2026-08-27
Notification use OS setting on mobile

jadehopepunk:
On mobile ensure that:

- In-app notification setting defaults to true
- Attempts to send notifications don't even check the in-app setting, only use OS setting
- The notification setting in the app doesn't display

Also, add better trace statements, containing the "[notification]" string for easier grepping.


guillemcordoba (approved):
I wasn't in the discussion earlier that agreed on this, this is a clear departure from Signal's UX. If this is the agreed upon change the diff looks good! 

I think Claude is right that e2e tests on mobile will fail after this, I'd change them as well.

PR #545, opened 2026-08-27
Disable message actions in read-only chats

guillemcordoba:
## Summary

After leaving a group you could still swipe-to-reply, react, and edit messages — actions that publish operations into a chat you can no longer write to. This PR introduces a read-only notion for chats at the store layer and disables those actions across every surface.

- **`MessagesStore.readOnly`**: a required, public `() => ReactivePromise<boolean>` constructor dependency. `GroupChatStore` wires it to `!me().member`; `DirectChatStore` passes a constant `false`.
- Components read it from the `messages-store` context (no prop drilling):
  - `SwipeToReply` — the gesture goes inert (`useReactiveValue`, since the gate sits in a touch handler).
  - `MessageActionsMenu` — Reply and Edit items are hidden; Copy and Delete (local actions) remain.
  - `MessageHoverToolbar` — the React and Reply buttons are hidden; the ⋯ menu remains.
  - `MessageActionsOverlay` — the long-press spotlight no longer shows the quick-reaction bar (`SpotlightOverlay.above` is now optional, and the ensemble bump handles the missing bar).
  - `ReactionsSheet` — tap-to-remove your own reaction is disabled; reaction chips stay visible and the who-reacted sheet stays browsable.

Note: enforcement is UI-level; rejecting non-member ops in the backend is separate work.

### DIFF
--- a/ui/src/lib/components/messages/MessageActionsMenu.svelte
+++ b/ui/src/lib/components/messages/MessageActionsMenu.svelte
@@ -7,7 +7,9 @@
 		mdiReply,
 	} from '@mdi/js';
 	import { List } from 'konsta/svelte';
-	import type { DeviceId, Message } from 'dash-chat-stores';
+	import { getContext } from 'svelte';
+	import type { DeviceId, Message, MessagesStore } from 'dash-chat-stores';
+	import { useReactivePromise } from '$lib/stores/use-signal';
 	import { canEditMessage } from './message-helpers';
 	import ListAction from '$lib/components/navigation/ListAction.svelte';
 	import DeleteMessageDialog from './DeleteMessageDialog.svelte';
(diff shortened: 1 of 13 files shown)


ui/src/lib/components/messages/MessageActionsMenu.svelte
```diff
@@ -38,28 +40,33 @@
...
-	{#if onReply}
-		<ListAction
-			title={m.reply()}
-			icon={mdiReply}
-			onClick={onReply}
-			data-testid="message-action-reply"
-		/>
-	{/if}
+	{#await $readOnly then readOnly}
+		{#if canEdit && !readOnly}
```
jadehopepunk: Maybe wrap this while block of two list actions with it's own {#if !readonly}. That way, readers don't need to double check the each of the if-statements to understand

jadehopepunk (approved):
Looks great

PR #546, opened 2026-08-27
Always fully sync topics with push server

maackle:


guillemcordoba (approved):

maackle:
We're not going to use this for now, it's only useful once we start implementing "gracefully leave group"

PR #547, opened 2026-08-28
Voice messages opus transcoding

jadehopepunk:
Transcode all voice messages to Ogg Opus.


crates/opus-transcode/src/input/resample.rs
```diff
@@ -0,0 +1,34 @@
+//! Resampling: converting audio from one sample rate to another by estimating
+//! the wave's value at the new sample instants (which fall between the original
+//! samples). Needed because Opus only encodes a fixed set of input rates.
+
+/// Sample rates Opus can encode natively; anything else is resampled to 48kHz.
+const OPUS_RATES: [u32; 5] = [8_000, 12_000, 16_000, 24_000, 48_000];
```
maackle: Seems that anything else should be down/upsampled to the nearest rate.

But actually we probably just shouldn't support resampling, because we have control over the input sample rate and we should just make sure it's one of these and panic if not.

crates/opus-transcode/src/opus/decode.rs
```diff
@@ -0,0 +1,85 @@
+//! Decode an Ogg/Opus stream back to PCM wrapped in a WAV container, so a
+//! webview `<audio>` element can play it without native Opus support.
+
+use std::io::Cursor;
+
+use anyhow::{Context, Result};
+
+/// Opus always decodes at 48kHz.
+const OPUS_DECODE_RATE: u32 = 48_000;
```
maackle, line 9: We can experiment with this. If we're just targeting speech, 48k is overkill. Speech only uses frequencies up to about 8kHz, meaning 16kHz should be sufficient. For other sounds including music though 48k is appropriate

crates/opus-transcode/src/visualisation.rs
```diff
@@ -0,0 +1,48 @@
+//! Visualisation: reduce PCM to a small array of amplitude bars for drawing the
+//! voice-message waveform in the UI (heights trace loudness over time).
+
+/// Number of amplitude bars in the returned waveform.
+pub(crate) const WAVEFORM_BARS: usize = 48;
+
+/// Reduce PCM to `WAVEFORM_BARS` peak amplitudes in `0..=255`, with the loudest
+/// bar mapped to 255 so quiet recordings still fill the waveform.
+pub(crate) fn compute_waveform(pcm: &[i16]) -> Vec<u8> {
```
maackle: One thing to look out for here is if there is a "pop" at the beginning of the audio, that will dwarf the rest of the waveform down to nothing. Might be a smarter waveform function, like taking the average (or RMS) instead of max per bucket. Just a note in case we see bad waveforms.

crates/opus-transcode/tests/transcode.rs
```diff
@@ -0,0 +1,136 @@
+use std::io::Cursor;
+
+use opus_transcode::{decode_opus_to_wav, transcode_to_opus};
+
+/// The waveform bar count is an internal display constant; assert against the
+/// known value rather than reaching into the crate's private module.
+const WAVEFORM_BARS: usize = 48;
```
maackle: duplicated in visualisation.rs

maackle (commented):
looks good, just some notes about audio stuff

crates/opus-transcode/src/input/resample.rs
jadehopepunk: Ok, this is somewhat addressed. 

I now always resample sample to 16 kHZ (and it sounds fine)

On desktop, we *try* and record at 16 kHZ, although we may get overwritten by the OS, hence the need to resample anyway. If we did record at 16 kHZ, it's an early-out no-op.

On mobile, the plugin we're using is actually currently throwing away the setting we're giving it, and recording at 41 kHZ. This needs a fix, but either we need to submit a PR to that plugin or fork it. I'd like to do it properly this time, and submit a PR. I'll do that later.

Since our re-sampler is now often downsampling, apparently the previous linear algorithm is insufficient. I've pulled in a crate to do a better job of that.

crates/opus-transcode/src/visualisation.rs
jadehopepunk: Good catch. This is addressed now, and the waveforms look much better.

crates/opus-transcode/tests/transcode.rs
jadehopepunk: Fixed

maackle (approved):

PR #550, opened 2026-08-29
Add 4th status indicator

maackle:
For the first minute after a message is sent, it shows a spinning dotted circle with no arrow.
After a minute it switches to "trying to send", a stationary dotted circle around an arrow.

Also updates the existing svgs to new versions.


packages/stores/src/message-acks/display-delivery-status.ts
```diff
@@ -0,0 +1,25 @@
+import { MessageDeliveryStatus } from '../types';
```
guillemcordoba: Seems to me like this needs to live in the ui package rather than the store package. Unit tests can also live there.

ui/src/lib/components/messages/MessageStatusIndicator.svelte
```diff
@@ -1,38 +1,63 @@
...
 	}
 
 	const props: Props = $props();
+
+	let now = $state(Date.now());
+	const displayStatus = $derived(
+		displayDeliveryStatus(props.status, props.timestamp, now),
+	);
+
+	$effect(() => {
```
guillemcordoba: Use the special `withinWindow()` time utility to reactively update the UI based on time.

e2e-tests/specs/offline-transition-ux.spec.ts
```diff
@@ -109,15 +109,21 @@ describe('Offline UX', () => {
...
+			// downgrade, so it must be visible right after the send.
 			await agent1.waitUntil(
 				async () => {
 					const status = await agent1.directChatPage.lastMessageStatus();
-					return status === 'sending';
+					return status === 'unsent';
 				},
 				{ timeout: 5_000 },
 			);
+			await agent1.waitUntil(async () => {
```
guillemcordoba, line 128: Default test timeout will fail before this waitUntil passes, see the first Claude comment in the PR.

guillemcordoba (approved):
Looking good! Some changes I'd like to see before merging.

PR #557, opened 2026-09-09
Spawn a task for each mailbox poll

maackle:


crates/mailbox-client/src/manager.rs
```diff
@@ -461,7 +477,18 @@ where
...
+                            }
+                            let manager = manager.clone();
+                            tokio::spawn(async move {
+                                manager.poll_mailbox(&id).await;
+                                manager.end_poll(&id).await;
+                                // The mailbox is schedulable again, so the loop
+                                // must re-evaluate rather than sleep on a wait
+                                // computed while it was in flight.
+                                manager.trigger_poll_loop();
+                            });
```
maackle: should probably just move the task spawn into `poll_mailbox`.

maackle:
This is ready to review, and the PR comment about wakeups being dropped is addressed in #560. I didn't put it in this PR because it works better to build that on top of the merge with #558, because both touched the "pending sync" state in a way that had to be reconciled first.

guillemcordoba (approved):
🥳

PR #559, opened 2026-09-10
Switch mDNS stack from mdns-sd to swarm-discovery

guillemcordoba:
Rewrite of the mDNS stack to discover local hubs from mdns-sd to swarm-discovery to align it to iroh's own stack. 

The discovery and announcements code in the `local-hub-discovery` crate have been tested against real life Android devices and iterated until they emit Found/Lost events when we can be confident the mailboxes manager will be able to register/unregister them, in the least amount of time possible.


crates/local-hub-discovery/src/discovery.rs
```diff
@@ -0,0 +1,452 @@
...
+                return;
+            }
+        }
+    }
+}
+
+/// A swarm-discovery browser on the current interfaces, forwarding every
+/// sighting to `sightings`; fails if binding the multicast socket does.
+fn spawn_browser(
+    sightings: &UnboundedSender<Sighting>,
```
maackle: Why unbounded? Maybe add the reason why there's confidence in not needing backpressure. Or if not at least make it bounded and use try_send to drop overflow.

e2e-tests/helpers/fuzz/checks.ts
```diff
@@ -27,6 +27,14 @@ import type {
  *  user reads the app as slow or broken. */
 export const DISCOVERY_MS = 2_000;
 
+/** What a hub that stopped gets before it has to be off the chip. Longer than
+ *  [`DISCOVERY_MS`] because nothing goes on the wire when a hub goes away —
+ *  swarm-discovery neither sends a goodbye nor reads one — so a phone only
+ *  notices once the hub ages out of its swarm: 3τ of silence (2.1s at the
+ *  interactive cadence, for up to two hubs) checked on a ~0.86s sweep, then the
+ *  app's own unregister. 4s clears that worst case with margin. */
+export const DEPARTURE_MS = 4_000;
```
maackle, line 34: I wonder where these values come from: 

"3τ of silence (2.1s at the interactive cadence, for up to two hubs) checked on a ~0.86s sweep"

and if they can be included in this as a formula in case they change.

maackle (approved):

maackle (approved):
looks pretty good, I can't say I understand all the changes but I commented on what I could. I'm also pushing a few minor comment changes.

e2e-tests/helpers/fuzz/checks.ts
guillemcordoba: You're right, those numbers came from actual runs with the fuzz testing search, but I don't like having them here. What we want here is a top down constraint on how we want the system to behave, not custom-made explanations or constraints that fit _how_ the mechanism works. You're right to flag this, I'll remove the numbers from the comment altogether.

PR #558, opened 2026-09-10
Mailbox "fast push" when authoring

maackle:

PR #560, opened 2026-09-10
Add notion of "pending wakeup", unified with two forms of "pending poll"

maackle:
#557 highlighted the possibility of a wakeup request being swallowed by an in-flight poll. This solves that by allowing wakeups to be pending too. There are now 3 distinct requests that can be pending during an in-flight poll, with logic about how multiple requests get combined.

This PR is best read once both #557 and #558 are merged, because this is built upon a merge of the two.


maackle:
folded this into https://github.com/dash-chat/dash-chat/pull/565

PR #561, opened 2026-09-10
"Probe" mailboxes instead of waking them on network/background changes, and drop local mailboxes earlier

maackle:
Adds the concept of "probing", a sibling of "waking". Probing simply makes a single request, and on success, sets status to Active, but otherwise doesn't change status. Previously, we only had "waking", which sets all mailboxes to Active, which is highly inappropriate in most contexts. Waking should only happen when there is good evidence that the mailbox should be active now, and the only instance of that is for the cloud mailbox when a push notification arrives.

Additionally, when a local mailbox moves to Stopped status, we can deregister it immediately instead of waiting for the mDNS TTL to expire. If it comes back, it will re-announce itself over mDNS, which will allow us to re-register as Active.


maackle:
folded this into https://github.com/dash-chat/dash-chat/pull/565

PR #565, opened 2026-09-11
Mailbox polling improvements and simplifications

maackle:
The actual improvement goal of this work is to handle wakeup/network change more reasonably. Previously those chnages would cause all mailboxes to "wake up" to the Active status, which often just causes thrashing when the mailbox truly is not available. Now we have a new notion of "probing", alongside "wakeup", which immediately polls a mailbox *without* changing its status. If successful, the status will upgrade to Active, but otherwise it remains unchanged. On network change, now every mailbox gets probed once instead of fully woken up.

There are now 3 distinct request types that can be pending during an in-flight poll, with logic about how multiple requests get combined:
- "poll" schedules a debounced normal sync round with the mailbox. Can be optionally narrowed to a set of topics.
- NEW: "probe" schedules an immediate sync, regardless of status, allowing the status to be upgraded if successful. This solves the problem of wanting to check on all mailboxes during a network change, without proactively changing the connection status in order to do so.
- "wakeup" immediately sets the status to Active and schedules an immediate sync (only used with cloud mailbox on push notification receipt)

Additionally, when a local mailbox moves to Stopped status, we can now deregister it immediately instead of waiting for the mDNS TTL to expire. If it comes back online, it will re-announce itself over mDNS, which will allow us to re-register as Active. This way we don't have to keep polling even at the slower Stopped rate.

https://github.com/dash-chat/dash-chat/pull/557 highlighted the possibility of a wakeup request being swallowed by an in-flight poll. This PR solves that by allowing wakeups (and probes) to be pending too. All requests are now tracked.

Also lets the actual scheduled time be calculated from the actual pending request, rather than storing and updating the scheduled time separately.


src-tauri/src/mailbox.rs
```diff
@@ -107,6 +109,13 @@ async fn register_local_hub(node: &dashchat_node::Node, hub: DiscoveredHub) {
             .with_blob_reader(node.blob_reader()),
         )
         .await;
+    // A hub that stops answering is gone as far as we are concerned, whether or
+    // not its mDNS records have expired yet; the re-browse re-registers it if
+    // it comes back. Only the first registration arms this, since a re-browse
+    // re-registers every hub it still sees.
+    if newly_tracked {
+        node.mailboxes.unregister_on_stopped(&id).await;
```
guillemcordoba, line 108: Hum I found this surprising. Shouldn't this be an internal behavior of the Mailboxes manager itself? I can't think of a case in which you wouldn't do this.

guillemcordoba (approved):
Looks great! One minor question.

src-tauri/src/mailbox.rs
maackle: Hmm I don't understand what alternative you're pointing to, or which part surprises you. Maybe the code could be moved somewhere else, but the idea is that you only want to do this for local hubs, and this is the entry point for registering a local hub.

PR #567, opened 2026-09-14
Persist mailbox sync status across restarts

maackle:
## Summary

Mailboxes now persist their last known sync status (Active, Degraded, or Stopped) to the database. When a mailbox is re-registered after a restart or iOS node rebuild, it seeds its connection state from this persisted history instead of always starting as Active. This prevents aggressive re-polling of mailboxes that were backed off before the restart.

This is mainly a response to this TODO from a previous PR: https://github.com/dash-chat/dash-chat/pull/567/changes#r4008919752

## Implementation Details

- Persisted status is only recorded when it actually changes, avoiding unnecessary database writes.
- If status persistence fails, the error is logged but does not block operation (graceful degradation).
- The in-memory test backend (`SyncBackend::Mem`) also supports status storage for testing.
- All existing tests updated to pass the new `initial_status` parameter to `TrackedMailbox::new()`.
- New tests verify seeding behavior, persistence across restarts, and that unregister-on-stopped waits for a poll to confirm a seeded Stopped status.


src-tauri/src/node/app_node_manager.rs
```diff
@@ -136,12 +136,17 @@ impl AppNodeManager {
         // spawned its cloud-mailbox registration retry in `AppNode::new`
         // (cancelled by the slot's teardown on the next pause), and the slot
         // bumps the generation on a fresh build so forwarders re-bind.
-        // TODO: a rebuilt node starts every mailbox Active with no history, so
-        // a hub that was Stopped before backgrounding is polled aggressively
-        // again. Seeding the rebuilt tracker from the persisted sync tracker's
-        // last known status would let resume probe here like Android does.
```
maackle, line 142: This is what was addressed

guillemcordoba (approved):

PR #582, opened 2026-09-16
Throttled iroh-blobs bandwidth for testing

maackle:
Allows us to slow down blob transfer to test partial blob transfer, download progress, and the like


maackle:
@guillemcordoba @jadehopepunk maybe you can review this separately later, I need it in the base branch so I'm merging now

PR #587, opened 2026-09-17
Make p2panda-node event channel unbounded to eliminate deadlock

maackle:
It was actually possible for this channel to overflow and cause deadlock, which became a source of flakiness in at least one unit test. Now it's unbounded.


maackle:
@sandreae I wonder if you agree with this, including you in case you are using a bounded channel in other actor implementations and have some experience with the same.

guillemcordoba (approved):

PR #589, opened 2026-09-19
Blob download progress on attachments, pull-based

maackle:
## Summary

A rewrite of #568 from `develop`, keeping its intent and the edge cases its four review rounds surfaced, on a design that has less to go wrong. Photos, files and voice notes whose blob has not arrived yet show a progress ring with the bytes present locally, flip to a retry affordance when the byte count stops advancing for 10s, and return to the plain attachment once the blob is complete.

## Design

**Pull, not push.** `get_blob_progress(hashes)` reads each hash's bitfield off the iroh-blobs store in one `observe` call and reports `{hash, bytes, complete}`. `BlobStore` polls it for the blobs that have a live subscription (i.e. are on screen), all of them in one request every 500ms, with an immediate first tick on mount, and only while any of them is incomplete. A blob leaves the poll once complete. There are no observer tasks, no notification variant, no Tauri event and no listeners, so nothing can leak, race, back-pressure operations, or miss a completion.

**The `` keeps mounting while the blob downloads.** Its request through the `irohblob://` scheme handler is what asks the node to fetch the blob now; the ring overlays it. If the handler's 30s timeout hits mid-download the ring stays instead of the error icon, and the image reloads on its own once the blob completes.

**Unknown is not zero.** The relay is pending until the first snapshot resolves, so an already-local attachment never flashes a ring (cold start, iOS resume), and a failed progress read leaves the image rendering as it does today.

**Trackers outlive their subscription.** `BlobStore` keeps one `BlobProgressTracker` per blob asked about, so an attachment scrolled away and back keeps its stall clock and reports its last state at once; a tracker is pruned once its blob is complete and nothing shows it. Components subscribe through `useBlobProgress` at init, and an effect follows `hash`/`visible` from there.

**One on-demand fetch per hash.** `BlobSync::fetch_now` claims an in-flight set, so a tap on a stalled photo and the image request it triggers share one attempt instead of stacking 30s loops.

**Stalls are judged at snapshot time**, no timers to run or dispose. A stalled photo's tap re-requests the image; a stalled file or voice note's tap runs its normal action (save, play), which fetches through the handler. A tap while still downloading is answered with a "Still downloading…" toast.

**Test-only pause.** `set_blob_fetch_paused` (behind `e2e-tests` / `testing`) makes every fetch attempt, background loop and on-demand alike, give up at once, so a spec can observe the downloading state.

## Follow-up (not in this PR, and this PR is not meant to merge without it being planned)

Tracked in #592: the throttled-mailbox coverage from #568 is deferred so the mailbox-server changes can be reviewed on their own (test-only blob throttle on the mailbox, the e2e mailbox actually serving blobs to e2e apps, and the spec that watches progress climb, hold and resume). Nothing in this diff refers to it; the issue is the marker.

## Tests

- Rust: unit tests for `local_progress` against a bare store (unknown hash reports zero and incomplete, stored blob reports its size and complete), and a two-node integration test asserting snapshots never decrease and end complete at the blob's size.
- Stores: `BlobProgressTracker` and `BlobStore` suites, the latter with a fake client (unknown until the first snapshot, one batched request, complete and unsubscribed blobs leave the poll, publish only on change, a failed poll keeps the state unknown and retries, retry clears a stall, a re-subscribed blob reports its last state at once and keeps its stall clock).
- Rust: the in-flight guard admits a hash once until dropped.
- E2E: photo held at `0 B /` until fetching resumes, stalled photo recovers on tap without opening the lightbox, held file shows the ring and answers a tap with the toast, and a new voice-note spec for the same.

## Verified

`cargo nextest` on the node crate, stores tests, `svelte-check`, e2e `tsc`, `just format`.

Visual pass on the e2e build (two desktop agents, receiver's fetching paused with `set_blob_fetch_paused`, driven over the embedded WebDriver): the photo cell shows the indeterminate ring with its byte pill, then the stalled ring with the retry glyph; the file row and voice button show theirs at 32px and 36px; the reply-quote thumbnail and the lightbox filmstrip show the 20px ring, which reads as a small spinner or retry badge; the lightbox centres the ring in the viewport, which is where it belongs while there is no photo to letterbox. Resuming fetching cleared the rings on every surface that still had an image request in flight, and a tap on a photo whose request had timed out recovered it without opening the lightbox.

https://claude.ai/code/session_01FUNbYmm45e45KcA38Yyd3m


maackle:
Round 2 addressed in b47b9a26, except item 2 which is left as is by decision (the observer-gated set stays far below 128 in practice; a client-side cap and failure back-off can come with the throttled-mailbox follow-up if it ever trips).

1. **Trackers outlive the subscription.** `BlobStore` keeps a `Map<Hash, BlobProgressTracker>` across relay teardown; the relay seeds its value from the tracker's last state on re-subscribe and only registers `publish`. A tracker is pruned at teardown once its blob is complete. Two new store tests cover the immediate re-report and the stall clock surviving an unsubscribe. (Signalium also keeps the relay's `ReactivePromise` across deactivation, so a complete blob re-subscribed reports complete at once, which a third test pins.)
3. **One on-demand fetch per hash.** `BlobSync::fetch_now` claims an in-flight `HashSet` through a drop guard and returns `false` at once when the hash already has an attempt running; both `load_blob` and `fetch_blob_now` go through it. Unit test on the guard.
4. Comment on the `blob-image-loading` branch naming the invariant.
5. **`useReactiveValue` no longer runs inside `$derived`.** New `useBlobProgress(blobStore, () => hash, () => visible)` is called once at init and manages the subscription from an `$effect` with explicit `subscribe`/unsubscribe, resetting to unknown when hidden. All three attachment components use it. I kept the state in the parent rather than a child component because `BlobImage` needs it for `retryIfErrored` and the reload-on-complete effect, not only for the ring.

Smaller notes: ring stays indeterminate below 5% or with `total <= 0`; `everyDownloadStalled()` is false for an empty set; the voice button's ring wrapper is `aria-hidden` while stalled; the photo-tap behaviour is now stated in `retryIfErrored`'s doc comment; the held-photo assertion matches on the separator.

Process: tracking issue #592 filed and linked from the description, which no longer mentions the removed markers. Visual pass done on the e2e build with the receiver's fetching paused; details in the description's Verified section. One thing noticed on the way, not from this diff: the desktop settings gear renders above the lightbox overlay.

